Issue #17, opened 2020-02-06
MangoHUD does not work on ZorinOS 15.1


canalocaradoti:
Hello, how are you? First of all I would like to congratulate you for the work and I hope it will evolve even more and thus be a reference in the Linux world.

Well, I'm an editor on the Brazilian blog Diolinux and through the GamingOnLinux blog, I saw your tool and found it interesting and went to test it. But unfortunately on ZorinOS 15.1 Core (Gnome) and Lite (XFCE), MangoHUD is not working, I ran both the .sh script and tried to compile it and it was not. In theory ZorinOS uses the Ubuntu 18.04 base, and in some packages it can use the 18.04.1 base or the 18.04.3 base. I also saw that the meson is in version 0.47. I have a hybrid laptop from Dell (Intel + Nvidia), with a GTX 960M and using the 440.44 drivers. I wonder if you have any procedure to do or something like that. I am available to help.

Att
Ricardo

flightlessmango:
Was there any errors when you tried to build mangohud?
What command did you use to test mangohud with?

For basic I usually recommend downloading the vulkan-tools package and running `MANGOHUD=1 vkcube`

telans:
> I ran both the .sh script and tried to compile it and it was not.

Failed to compile? What was the error message?

canalocaradoti:
So, I followed the recommended procedures.  There was no error when I ran install.sh or when I ran ./build.sh install.  I also tried the Vulkan test and nothing, so much so that I have the vulkan-tools package installed.

telans:
Does `$HOME/.local/MangoHud/libMangoHud.so` exist? Same with `$HOME/.local/share/villain/implicit_layer.d/mangohud64.json`

canalocaradoti:
@telans here in my ZorinOS, there are these paths : /home/ricardo/.local/share/vulkan/implicit_layer.d/ with the mangohud64.json and /home/ricardo/.local/share/MangoHud/ with the libs libMangoHud.so and  libMangoHud32.so. I'm sending a print of when I run sh with and without the sudo command.
![2020-02-06_15-36](https://user-images.githubusercontent.com/41804119/73967829-6bbea680-48f7-11ea-99c8-9d1fe6f033e8.png)

telans:
Then it's strange that it doesn't load. Do you have vulkan-icd loader installed? Or in other words, does DXVK work on your system?

canalocaradoti:
[...]

PR #20, opened 2020-02-08
Use dpkg-query -s to query for packages

guihkx:
By own experience, `dpkg-query -l` doesn't always return 1 if a package isn't installed.

`dpkg-query -s` seems more reliable. Source:

https://github.com/bitrise-io/bitrise/issues/433

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -44,7 +44,7 @@ dependencies() {
                 DEPS="{glibc-devel.i686,libstdc++-devel.i686,libX11-devel.i686}"
                 install
             ;;
-            *"buntu"|"Linux Mint"|"Debian")
+            *"buntu"|"Linux Mint"|"Debian"|"Zorin OS")
                 MANAGER_QUERY="dpkg-query -s"
                 MANAGER_INSTALL="apt install"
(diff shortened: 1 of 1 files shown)


guihkx:
I've also added Zorin OS to the script, as per discussion [here](https://github.com/flightlessmango/MangoHud/issues/17#issuecomment-583733884). Hope you guys don't mind.

imaami:
I don't mean to poop on your party, but I have to say the build script is pretty horrible. It's full of obvious landmines in the form of not quoting any variables - please read up on how shell variable expansion behaves when a variable's value contains spaces. And there are some weird things as well, like running a command, then echoing `$?` and comparing the output to `1` - really?

But again I apologize for being like this. Just annoyed right now. The actual problem with the script is trying to do everything a package manager does, using sudo to do it without even stopping to ask, copying stuff to hardcoded paths, etc., and also trying to do what the build system (meson) is supposed to already accomplish.

It's just a giant do-everything-badly overreach of a script. You use meson, so don't put the install logic in some script (unless meson can't do it at all, in which case: why?). And when I'm not logged in as root, just trying to build a fairly small C/C++ project, you're not supposed to launch a system-wide package install command using sudo or whatever! That's not OK. Your build script or system should tell me if I'm missing programs or tools, not try to actually install packages it _thinks_ I need.

There are more examples in the script that are also worth mentioning, but I'm kind of burned out from this rant already so I'll pass.

telans:
If you want to improve it, feel free to submit a pull request

PR #25, opened 2020-02-11
Add x11/wayland feature switches

DanySpin97:


flightlessmango:
I think this was in before but removed because mangohud does not work on wayland for some reason

telans:
segfaults for me

meson_options.txt
```diff
@@ -1 +1,3 @@
 option('glibcxx_asserts', type : 'boolean', value : false)
+option('with_x11', type : 'feature', value : 'enabled')
```
jackun: Can't this be just a boolean?

jackun:
Atleast hotkeys stuff needs to be \#ifdef'ed out if disabling x11

jackun:
Merged manually.

PR #41, opened 2020-02-17
Add Solus support

ikeycode:
Full rationale is given per commit, but to overview:

 - Support stateless layout for the `os-release` file location
 - Use eopkg pythonic APIs to verify presence of 32bit build dependencies and `system.devel` component. Limitations are explained in the individual commit but this chiefly comes down to exit codes. Tested as fully working on Solus and non-interfering with other distros.


ikeycode:
Screenshot from 2020-02-17 11-22-14

PR #83, opened 2020-03-17
Use wrap for Dear ImGui

stephanlachnit:
This PR aims to replace the Dear ImGui git submodule with a proper [wrap dependency](https://mesonbuild.com/Wrap-dependency-system-manual.html). This has some advantages:
- Updating only requires changing a file
- Works without a git directory (also doesn't require to init the submodules)

However, it does require a second repository where you handle the `meson.build` file. Right now, I created one [here](https://github.com/stephanlachnit/dearimgui-wrap). I would give you the ownership of the repository.

This is still a draft because of three reasons:
- Right now you use the experimental tables branch, wrap however only works with releases so that would have to wait until the next ImGui release.
- The repository is still in my ownership (should be in the hand of someone who actually changes the version).
- I didn't remove the git submodule yet since removing submodules will be messed up when rebasing.

I also updated the name of the dependency variable to follow meson's recommendation of `<project-name>_dep`. I am also thinking about doing the same for the Vulkan-Headers.


stephanlachnit:
Just put the Vulkan-Headers in a wrap subproject as well. Except the Dear ImGui tables stuff it looks like it works just fine.

jackun:
I merged vulkanheaders.wrap part to `develop`. Couldn't wait :P

stephanlachnit:
Nice! I just invited you and @flightlessmango to both wrap repos :)

stephanlachnit:
@jackun I updated the `build-source.sh` script for wrap, so be aware of that when merging `develop`. I also updated the Vulkan Headers to `1.2.137` :)

stephanlachnit:
Closed in favor of #153.

PR #95, opened 2020-03-22
Updated ManogoHud.conf with the recent values

### LINKED ISSUE #92: update master mangohud.conf
new variables needed to be added to config
https://github.com/flightlessmango/MangoHud/blob/master/bin/MangoHud.conf

> reload_cfg= | Change keybind for reloading the config
> 
> timetime_format=%T | Displays local time. See std::put_time for formatting help.
> 
> text_color | #ffffff
> gpu_color | #2e9762
> cpu_color | #2e97cb
> vram_color | #ad64c1
> ram_color | #c26693
> engine_color | #eb5b5b
> io_color | #a491d3
> frametime_color | #00ff00
> background_color | #020202
> Change default colors: gpu_color=RRGGBB
> 
> alpha | Set the opacity of all text and frametime graph 0.0-1.0
> 
> background_alpha | Set the opacity of the background 0.0-1.0
> 
> read_cfg | Add to MANGOHUD_CONFIG as first parameter to also load config file. Otherwise only MANGOHUD_CONFIG parameters are used.
> 
> output_file | Define name and location of the output file (Required for logging)
> 
> font_file | Change default font (set location to .TTF/.OTF file )
> 
> log_duration | Set amount of time the logging will run for (in seconds)
> 
> vsync
> gl_vsync | Set vsync for OpenGL or Vulkan
> 
> io_write
> io_read|IO read and write for the app, not system.

i added these to my mangohud.conf
i figure it would be easier if everyone had them by default

Faalagorn:
Fixes #92

### DIFF
--- a/bin/MangoHud.conf
+++ b/bin/MangoHud.conf
@@ -70,15 +70,15 @@ background_alpha=0.5
 # alpha=
 
 ### Color customization
-# text_color	#ffffff #ffffff
-# gpu_color	#2e9762 #2e9762
-# cpu_color	#2e97cb #2e97cb
-# vram_color	#ad64c1 #ad64c1
-# ram_color	#c26693 #c26693
-# engine_color	#eb5b5b #eb5b5b
-# io_color	#a491d3 #a491d3
-# frametime_color	#00ff00 #00ff00
-# background_color	#020202 #020202
+# text_color=FFFFFF
+# gpu_color=2E9762
+# cpu_color=2E97CB
+# vram_color=AD64C1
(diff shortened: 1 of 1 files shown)


flightlessmango:
I don't understand why the colors are written twice and I don't think it works to write `#` in the color.
Otherwise it looks good

Faalagorn:
Thanks for pointing that out! I copied that section from the comment without thinking. I now fixed it and changed it to big letters to match that of crosshair and readme.

PR #119, opened 2020-04-08
Write log in a memory buffer and periodically write it in disk

andrealmeid:
Hello,

This PR improves the performance of logging by writing data in a memory buffer and periodically writing it to disk, to avoid losing data in case of interruption. The time interval of writes can be set using the environment variable `LOG_WRITE_INTERVAL` and it is set as 1 second by default. Writing to the file for each line can hurt performance since it will trigger a disk write and this action can be blocked.

To test this, I have setup the following environment:  
- `LOG_WRITE_INTERVAL = 1 sec`
- `LOG_PERIOD = 10`
- `output_file`: a file in a USB flash drive
- Used the benchmark tool from "Metro: Last Light Redux". Owners of this game can find this tool in `.local/share/Steam/steamapps/common/Metro Last Light Redux/benchmark.sh`.

This was the result:

buffer_disk

I started the logging at the same time, and it should have the same duration. However, as can be seen, the "Write on disk" test took long to run, since a lag was added between the frame draws, making the a horizontal scaling (like when you multiply _x_ by a constant).

It also improves the indentation and don't call sleep if the log duration is over.

Thanks!

### DIFF
--- a/src/logging.h
+++ b/src/logging.h
@@ -60,8 +60,8 @@ void *logging(void *params_void){
 
     if ((elapsedLog) >= params->log_duration * 1000000 && params->log_duration)
       loggingOn = false;
-
-    this_thread::sleep_for(chrono::milliseconds(log_period));
+    else
+      this_thread::sleep_for(chrono::milliseconds(log_period));
   }
 
   writeFile(params->output_file + date);
--- a/src/logging.h
+++ b/src/logging.h
@@ -10,16 +10,16 @@ using namespace std;
 
 string os, cpu, gpu, ram, kernel, driver;
 bool sysInfoFetched = false;
-int gpuLoadLog = 0, cpuLoadLog = 0, log_period = 0;
+int gpuLoadLog = 0, cpuLoadLog = 0, log_period = 0, elapsedLog;
 
 struct logData{
   double fps;
-  double cpu;
-  double gpu;
-  double previous;
+  int cpu;
+  int gpu;
+  int previous;
 };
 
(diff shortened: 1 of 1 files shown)


flightlessmango:
I can imagine this possibly being an issue. I have however not encountered these issues with log duration, I normally log to a ramdisk disk so I might be avoiding it like that

I'm thinking the variable `LOG_WRITE_INTERVAL` is not necessary, just write to memory until done and then paste the whole thing

andrealmeid:
> I'm thinking the variable `LOG_WRITE_INTERVAL` is not necessary, just write to memory until done and then paste the whole thing

This was my first implementation design, but while testing I saw that when the game process was killed/interrupted (like when using Ctrl + C in terminal), nothing was written to the file. To prevent data loss, I added this option. 

I could add a rule that `if (LOG_WRITE_INTERVAL == 0)`, write only once when the log is done, what do you think?

flightlessmango:
Yes the log would be lost, but what use is a broken log? I can't quite see the use case for it. If my game crashes mid benchmark, I'm doing a new log

andrealmeid:
Hello @flightlessmango, could you have a look in the changes again?

I removed the log duration as suggested, but keep in mind that the log will also be lost in the case you manually closes the game while the log is happening. This occurs because `loggingOn` never is `false` is this case. I tried to find some function that is called when the HUD is being shutting down, but I wasn't able to find it. If there's such function, just adding `loggingOn = false;` will solve this case.

Thanks :)

Issue #118, opened 2020-04-08
sudo is mandatory


TheEvilSkeleton:
Greetings, so I have encountered an issue where it is required to use sudo. I am a doas user and do not have sudo installed, so I get this issue: 
![image](https://user-images.githubusercontent.com/50847364/78734597-2ab73280-7938-11ea-91d3-77675237afa3.png)
Note that installing sudo made it work just fine, but it is still a bit annoying.

PR #123, opened 2020-04-12
Add doas support

Newbytee:
Adds support for using doas if sudo is not available. Closes https://github.com/flightlessmango/MangoHud/issues/118


Faalagorn:
I'm not using the build script myself, but do you think it's possible to add `su -c` while at it? Arch PKGBUILDs support it when sudo is not installed.

Newbytee:
> I'm not using the build script myself, but do you think it's possible to add `su -c` while at it? Arch PKGBUILDs support it when sudo is not installed.

I was considering this, but ultimately decided against it as `su -c` requires quotation marks. Bit less simple to implement. I'll look into it though, as it definitely would be nice to neither depend on sudo nor doas.

Newbytee:
Decided upon this: I'll work on support for using `su` directly, but it will be a separate pull request.

PR #131, opened 2020-04-18
Blacklist: Add OpenGL and fix `mangohud steam`

Sporif:
This adds the ability to run `mangohud steam` and have it work with all games by default. That requires:
 - adding the blacklist functionality to OpenGL
 - not hooking dlsym as this breaks steamwebhelper and doesn't seem to be required by any games I've tried
 - fixing `LD_PRELOAD` that is currently unset and never reset.


jackun:
> not hooking dlsym as this breaks steamwebhelper and doesn't seem to be required by any games I've tried

Try Live for Speed, wined3d doesn't get hooked, for me at least.

Sporif:
That's odd, other wined3d games (such as Arkham Asylum with `PROTON_USE_WINED3D=1`) do get hooked.

PR #150, opened 2020-05-03
Use gzip in build-source.sh

stephanlachnit:
I just noticed I messed up the script - it didn't use gzip (or any other compression). Debian doesn't accept non-compressed tarballs, which means the uploaded ones are useless to me.
Can you run the tar commands by hand on the v0.3.5 tag (the script on that tag has to be unchanged else there will be diff errors), and publish the tarballs? I'm really sorry for this.

I also removed an old entry in the gitignore I forgot to remove in #148.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,7 +6,6 @@ MangoHud*.tar.gz
 pkg
 mangohud*.tar.*
 lib32-mangohud*.tar.*
-v*.tar.gz
 
 # Prerequisites
 *.d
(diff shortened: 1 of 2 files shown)


stephanlachnit:
@jackun thanks, but I think you didn't had the Dear ImGui submodule downloaded, because `modules/ImGui/src/` is empty in the tarball. Can you re-upload it with that?

Honestly I can't wait for the next Dear ImGui release so that #83 can be merged (no more git submodules).

jackun:
Oof, ok. Github upload is weirdly slow, especially with those gifs.

PR #151, opened 2020-05-03
add metainfo file

stephanlachnit:
I created a metainfo file for MangoHud (using https://www.freedesktop.org/software/appstream/metainfocreator/#/consoleapp). This makes MangoHud showing up in Gnome Software, KDE Discover etc. if it's available in the distro's repo.

It isn't perfect though, I'm not sure whether the "screenshots" will be properly shown (normally you would want to use a 16:9 png), but I guess it should be easy to test it, and it doesn't have an icon yet.

Even though this is only a console application, I think an icon is nice for the metainfo page. The icon should be an svg. I think @flightlessmango's user picture would like nice, but ofc that's not my decision. I prepared everything for the icon.


flightlessmango:
I don't fully understand what this is, but I guess it wouldn't hurt. My profile picture would be fine

stephanlachnit:
> I don't fully understand what this is, but I guess it wouldn't hurt. My profile picture would be fine

Basically it's meta information for App Stores, making MangoHud better visible and easier to install for Linux newbies. It's ready to merge as is, the picture can be added later.

Do you have the picture as an svg? That would be the easiest. It also should have a free license (CC-BY-SA 4.0 International for example is great for artwork, if you can choose a license).

flightlessmango:
The purest form I have is an `.ai` file, I'm not a wiz when it comes to image stuff tho so I'm not sure how to make it svg. I also don't know how to add a license

stephanlachnit:
So it's Adobe Illustrator, right? Then you should be able to export it via File->Export and then select svg somewhere (scalable vector graphics).

Since svg's don't have a license attribut, I would edit the file with a text editor and add SPDX/REUSE comment (xml style): https://reuse.software/spec/

flightlessmango:
I'm not the original creator and I do not have that program. The original would also need to be cropped and without the knowledge or program I'm not sure how to proceed

stephanlachnit:
> I'm not the original creator and I do not have that program. The original would also need to be cropped and without the knowledge or program I'm not sure how to proceed

`.ai` files can also be loaded with Inkscape (which is open source), there it can also be exported as svg. I haven't worked much with Inkscape, so I don't know how to crop either, but I'm sure it's possible in a not too complicated way.

But if you are not the original creator, are you even allowed to publish it with a CC license?

flightlessmango:
Yes it's a logo that I commissioned. I will have a look at inkscape

flightlessmango:
This should be an svg and should contain a CC license
[FlightlessMango.svg](https://cdn.discordapp.com/attachments/366023719199440908/777018315318886420/FlightlessMango.svg)

stephanlachnit:
Pushed with the svg. In KDE Discover the icon isn't shown however, I suspect because it is not a square.

Gnome Software currently only shows metainfo of type `desktop-application`, so sadly MangoHud isn't shown there.

data/meson.build
```diff
@@ -1,4 +1,34 @@
...
+              '--no-net',
+              '--pedantic',
+              metainfo_file]
+  )
+endif
+
+# Install metainfo file
+install_data(
+  metainfo_file,
+  install_dir: join_paths(datadir, 'metaino')
```
jackun, line 21: metaino? Should be metainfo?

stephanlachnit: Ah whoopsi yes you're right.

stephanlachnit:
@jackun @flightlessmango I rebased this against #630.

Right now, this sadly can't really be tested in with KDE Discover or Gnome Software (e.g. https://gitlab.gnome.org/GNOME/gnome-software/-/issues/1476), as they don't really check local files.

I would suggest to just merge this after #630 and then it should be much easier to test it.

jackun:
Ah github dun goof

PR #152, opened 2020-05-03
Add .editorconfig for automatically setting indentation formatting in…

jackun:
… supported IDEs or text editors

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,22 @@
+# See http://editorconfig.org to read about the EditorConfig format.
+# - Automatically supported by VS2017+ and most common IDE or text editors.
+# - For older VS2010 to VS2015, install https://marketplace.visualstudio.com/items?itemName=EditorConfigTeam.EditorConfig
+
+# top-most EditorConfig file
+root = true
+
+# Default settings:
+# Use 4 spaces as indentation
+[*]
+indent_style = space
+indent_size = 4
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[src/overlay*{cpp,h}]
+indent_size = 3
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -0,0 +1,22 @@
...
+# Default settings:
+# Use 4 spaces as indentation
+[*]
+indent_style = space
+indent_size = 4
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[src/overlay*{cpp,h}]
+indent_size = 3
```
stephanlachnit, line 17: Are you sure this shouldn't be four?

.editorconfig
```diff
@@ -0,0 +1,22 @@
...
+trim_trailing_whitespace = true
```
stephanlachnit, line 18: Also this whole section is kinda unnecessary since the behavior is already defined in `[*]`.

.editorconfig
jackun: Most of overlay.* etc are 3 spaces as it was in mesa vulkan overlay.

stephanlachnit: Ah whoops I just read `*{cpp,h}`, so nvm.

stephanlachnit:
Btw I would add an exception for Dear ImGui, else it could add newlines in there.

jackun:
> Btw I would add an exception for Dear ImGui, else it could add newlines in there.

It has its own .editorconfig and at least with KDevelop it seems to override MangoHud's settings as it has [root = true](https://stackoverflow.com/a/30539967)

stephanlachnit:
Ah whoops I missed that.

PR #153, opened 2020-05-04
Import Dear ImGui files as subproject

stephanlachnit:
This is an alternative to #83. It has the advantage that it can be used instantly, since it doesn't depend on any release tarballs. It also removes a lot the the license hassle (MangoHud's source tarball will only include the MIT licensed files, not some of the other licensed example inside Dear ImGui).
It does require manual upgrading though, meaning every file needs to be copied.

I actually think it's easier and better than #83, and way better than the git submodule.


stephanlachnit:
@jackun @flightlessmango any opinions on this? I updated it to make #225 redundant.

jackun:
Might as well for adding custom patches to ImGui. Waiting for text shadow/borders might take forever.

PR #158, opened 2020-05-05
Blacklist Additions

FaithLV:
This PR:
- Blacklists `LeagueClient` and `LeagueClientUxRender` which would crash with `MANGOHUD=1` 
- Replaces debug stream write which now always displays blacklisted processes when detected
- Introduces `reinstall` command to `build.sh` which bundles `build, package, install`

League of Legends now works with this PR - [benchmark](https://flightlessmango.com/games/2829/logs/252).

### DIFF
--- a/src/blacklist.cpp
+++ b/src/blacklist.cpp
@@ -39,7 +39,7 @@ static bool check_blacklisted() {
         "ffxivlauncher.exe",
         "ffxivlauncher64.exe",
         "LeagueClient.exe",
-        "wine-LeagueClientUxRender",
+        "LeagueClientUxRender.exe",
         "lutris"
     };
 
--- a/src/blacklist.cpp
+++ b/src/blacklist.cpp
@@ -38,6 +38,7 @@ static bool check_blacklisted() {
         "Steam.exe",
         "ffxivlauncher.exe",
         "ffxivlauncher64.exe",
+        "lutris"
     };
 
     std::string proc_name = get_proc_name();
--- a/src/blacklist.cpp
+++ b/src/blacklist.cpp
@@ -44,7 +44,7 @@ static bool check_blacklisted() {
 
     std::string proc_name = get_proc_name();
     bool blacklisted = std::find(blacklist.begin(), blacklist.end(), proc_name) != blacklist.end();
-    fprintf(stderr, "INFO: process %s is blacklisted in MangoHud %d\n", proc_name.c_str());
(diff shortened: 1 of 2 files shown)


src/blacklist.cpp
```diff
@@ -38,13 +38,14 @@ static bool check_blacklisted() {
...
+        "wine-LeagueClientUxRender",
+        "lutris"
     };
 
     std::string proc_name = get_proc_name();
     bool blacklisted = std::find(blacklist.begin(), blacklist.end(), proc_name) != blacklist.end();
-#ifndef NDEBUG
-    fprintf(stderr, "MANGOHUD: process %s is blacklisted: %d\n", proc_name.c_str(), blacklisted);
-#endif
+    fprintf(stderr, "INFO: process %s is blacklisted in MangoHud\n", proc_name.c_str());
```
jackun, line 48: You'd have to check `blacklisted` if it is actually blacklisted.

src/blacklist.cpp
```diff
@@ -38,13 +38,14 @@ static bool check_blacklisted() {
         "Steam.exe",
         "ffxivlauncher.exe",
         "ffxivlauncher64.exe",
+        "LeagueClient.exe",
+        "wine-LeagueClientUxRender",
```
jackun, line 42: This should probably be `LeagueClientUxRender.exe`

src/blacklist.cpp
```diff
@@ -38,13 +38,14 @@ static bool check_blacklisted() {
...
+        "lutris"
```
jackun, line 43: Are you sure? `lutris` is a python script and `proc_name` would be some variation of `python`

FaithLV: Yep! Log does show that `python3.8` was blacklisted, but only adding `lutris` allows lutris to launch.

jackun: Weird, does it show up in the log as "blacklisted"? Could be some utility app that lutris uses like glxinfo etc. Lutris seems to work ok for me either way :confused:

FaithLV: Yep, seems that segfault with lutris was not releated, removed from blacklist.

jackun:
Merged manually with some tweaks

FaithLV:
Could I get the commit in which this included? 
I want to see how it turned out in the end.

jackun:
In [develop](https://github.com/flightlessmango/MangoHud/commits/develop)

PR #159, opened 2020-05-06
Implement keybindings with multiple keys

FaithLV:
This PR:
- Implements keybinding configuration to support arrays of keys
- Adds `GNU Global` metadata to `.gitignore`
- Sets default keybind for `reload_cfg` to `LShift+F4`

In theory, should support any number of keys per a binding.
Space as a delimiter would be a good default, I think. 

These are my first contributions for a `C++` project, so I expect I'll have to iterate on it before merging.

### DIFF
--- a/src/keybinds.h
+++ b/src/keybinds.h
@@ -16,7 +16,7 @@ bool keys_are_pressed(std::vector<KeySym>& keys) {
         return false;
 
     char keys_return[32];
-    std::vector<bool> pressed;
+    size_t pressed = 0;
 
     g_x11->XQueryKeymap(get_xdisplay(), keys_return);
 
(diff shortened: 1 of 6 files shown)


src/keybinds.h
```diff
@@ -10,15 +10,32 @@ double elapsedF2, elapsedF12, elapsedReloadCfg;
...
 
 #ifdef HAVE_X11
-bool key_is_pressed(KeySym ks) {
+bool keys_are_pressed(std::vector<KeySym>& keys) {
 
     if (!init_x11())
         return false;
 
     char keys_return[32];
+    std::vector<bool> pressed;
```
jackun, line 19: Might as well be size_t if just use for counting

FaithLV: Replaced with `size_t`

jackun:
Tweaked version in 1afea91dab5a30bfd9e65968af038ceaf95d3c39

PR #161, opened 2020-05-07
get driver properties

DadSchoorse:
~~Here, checking for `VK_KHR_get_physical_device_properties2` turned out a little bit difficult, but it's so old that we can assume support for it.~~
only works on > 1.1 because the loader is stupid,


flightlessmango:
This has been manually merged into develop 848b2964778d66a7aeb7c1de1c6bce0c4c490e32

PR #197, opened 2020-06-02
Blacklist lutris

FaithLV:
Setting `MANGOHUD=1` globally causes lutris to segfault.
This PR adds lutris to blacklist.

Sometimes it doesn't segfault and for others it seems it never segfaults. We probably don't want MangoHud for lutris itself anyway.

PR #204, opened 2020-06-08
Change blacklist type

FaithLV:
This PR changes blacklist to contain `std::string` instead of `const char*` which fixes segfaults in `lutris` and `vulkaninfo`.

Also changes blacklist to be `std::vector` instead of an array as I think it helps with maintainability.

Discussion: https://github.com/flightlessmango/MangoHud/pull/197


jackun:
This is probably just curing a symptom not a cause

PR #207, opened 2020-06-11
build-source.sh: Add parent directory to source tarballs

akien-mga:
Also ensure that submodules have been initialized.

---

Most source distribution tarballs include a parent folder that matches the tarball's name (without extension), so this change makes this follow the same convention.

### DIFF
--- a/build-source.sh
+++ b/build-source.sh
@@ -1,10 +1,16 @@
 #!/bin/sh
 
 VERSION=$(git describe --tags --dirty=+)
+NAME=MangoHud-$VERSION-Source
 
 FILE_PATTERN="--exclude-vcs --exclude-vcs-ignores ."
 
+# Ensure that submodules are present
+git submodule update --init
+
+mkdir -p build
+
 # default version
-tar -czf MangoHud-$VERSION-Source.tar.gz $FILE_PATTERN
+tar -czf build/$NAME.tar.gz $FILE_PATTERN --transform "s,^\.,$NAME,"
 # DFSG compliant version, excludes NVML
-tar -czf MangoHud-$VERSION-Source-DFSG.tar.gz --exclude=include/nvml.h $FILE_PATTERN
+tar -czf build/$NAME-DFSG.tar.gz --exclude=include/nvml.h $FILE_PATTERN --transform "s,^\.,$NAME-DFSG,"


jackun:
I get

```
tar: .: file changed as we read it
tar: .: file changed as we read it
```
and `.` gets added too
```
./
MangoHud-v0.4.1+-Source/.editorconfig
...
```

akien-mga:
> I get
> 
> ```
> tar: .: file changed as we read it
> tar: .: file changed as we read it
> ```
> 
> and `.` gets added too
> 
> ```
> ./
> MangoHud-v0.4.1+-Source/.editorconfig
> ...
> ```

Yeah I'm not sure why this happens, but I can't find any good documentation on how to tar the cwd's files while adding a parent folder. I'll dig some more.



> and `.` gets added too
> 
> ```
> ./
> MangoHud-v0.4.1+-Source/.editorconfig
> ...
> ```

That was already the case before this patch, but I don't understand why the `--transform` regex doesn't properly change it to `MangoHud-v0.4.1+-Source/`.

akien-mga:
> tar: .: file changed as we read it

This seems to predate my changes, the same happens on `v0.4.1` when the tarballs are missing. I get `tar` doesn't like having a tarball appear in `.` when archiving `.` (even if it's then ignored due to `.gitignore` and `--exclude-vcs-ignores`.

To fix it, the tarballs should be generated somewhere else (`../`, or `build/` maybe).

> and `.` gets added too

I tweaked the `--transform` and now it works it seems.

akien-mga:
I pushed an amended commit that should address the above issues.

jackun:
Does tagging work on github :P 
@stephanlachnit  looks good?

stephanlachnit:
> Most source distribution tarballs include a parent folder that matches the tarball's name (without extension), so this change makes this follow the same convention.

Thanks! 

> Does tagging work on github :P
> @stephanlachnit looks good?

Looks good, haven't tested it though.
Tbh I want to change the archive creation anyway, using `git archive` instead of `tar` (something like [here](https://github.com/FeralInteractive/gamemode/blob/master/scripts/mkrelease.sh)) since it removes the gitignore hassle.

> Yeah I'm not sure why this happens, but I can't find any good documentation on how to tar the cwd's files while adding a parent folder. I'll dig some more.

This is because I wrote that file quickly so that it works, but it's not exactly best practice. The problem is, using `.` will pack _all_ files in that folder into the tarball, but we write the tarball to the directory as we do it, that's the reason for the warning.
Normally you would rather use something like `*` which doesn't produce the warning, but that doesn't include files starting with a `.`, which is problematic since we want an accurate copy of the git repo. This is why it's better to use `git archive` and then add imgui manually, like I did for gamemode (link above). I just didn't have the time to do it.

akien-mga:
Yeah I thought about changing things to use `git archive`, but it's a hassle that it won't include submodules (which is IMO the main missing Git feature that makes submodules a very subpar and problematic feature).

stephanlachnit:
You can include the submodule to the already created archive.

stephanlachnit:
See #216, should fix this (at least for the non-DFSG version).

akien-mga:
Thanks, that looks good to me. A slight nitpick would be that the DFSG version will not have the `-DFSG` suffix in its parent folder name, but it's not too important IMO.

I'll close this as superseded by #216.

stephanlachnit:
> Thanks, that looks good to me. A slight nitpick would be that the DFSG version will not have the `-DFSG` suffix in its parent folder name, but it's not too important IMO.
> 
> I'll close this as superseded by #216.

It would be possible (basically doing everything twice), but at least I don't need it for the Debian package, I guess other distros won't use the DFSG version anyway^^

akien-mga:
> It would be possible (basically doing everything twice), but at least I don't need it for the Debian package, I guess other distros won't use the DFSG version anyway^^

Well most other distros will also not want to include any nonfree content in their tarballs, so I'm using the DFSG one for Mageia's package. But yeah, I don't mind the absence of the `-DFSG` suffix.

PR #225, opened 2020-06-15
build-source.sh: only pack required imgui files

stephanlachnit:
@jackun @flightlessmango 
This is attempt to reduce the somewhat complicated copyright attribution for MangoHud ([here](https://salsa.debian.org/games-team/mangohud/-/blob/debian/master/debian/copyright) is my review) by removing all unneeded Dear ImGui files.
Just because I think it should be mentioned #153 also helps in this regard since it reduces the diffs between git repo and source tarball (which will be quite big with this commit).


stephanlachnit:
Closed in favor of #153.

Issue #224, opened 2020-06-15
CrossHair


remi75:
Hello and thank you again for mangohud :)

May we know why the cross-hair feature has been removed ?

Regards

flightlessmango:
Because it can be argued to be a cheat. My reasoning is that it would be very unfortunate if MangoHud got blacklisted as a cheat by anticheats simply on the basis of that one feature.
I think it's important to be clear with what MangoHud is and what it's used for. The crosshair feature muddies the water in that respect

remi75:
That is your software and opinion :) 
You are giving me the occasion to fork it and re-implement it ;)
At least, I'll try .

Regards.

flightlessmango:
I've contemplated making a separate project that adds crosshair etc, based on mangohud but doesn't have any of the other mangohud features

Thann:
Sorry for the necrobump, but I re-added the feature on my fork =]

https://github.com/Thann/MangoHud

spider3000:
> I've contemplated making a separate project that adds crosshair etc, based on mangohud but doesn't have any of the other mangohud features

Also apologies for the necrobump. Thanks for this amazing project. Out of curiousity, have you given this separate project more thought? Would love to see overlay tools like this in a well-maintained up-to-date project, that would be compatible with the newest Mangohud.

PR #228, opened 2020-06-17
Add configurable crosshair to overlay

swissChili:
Some games either do not have crosshairs at all (Insurgency), or do not have them enabled for certain weapons (like the AWP and Scout in CS:GO). This pull request adds the option to render a small circle in the middle of the screen, with a configurable size and color, usable as a crosshair.

I have added four configuration options that can be used to modify the appearance of the crosshair. It can also be disabled easily.

```sh
crosshair # enable the crosshair
crosshair_radius = 15 # radius (in pixels)
# ARGB8888
crosshair_color = DD22FF00 # color
crosshair_border = 2 # thickness in pixels
```

I don't know if this is within the scope of MangoHud, if not feel free to close the issue, but this was something that I found useful.


swissChili:
Here is an example of how it looks:

Crosshair

Faalagorn:
It was previously included, but was removed in 0.4.0, see also #224

Issue #195, opened 2020-06-02
0,1 & 1 % Low FPS


kgudalov:
Is it possible to implement a monitoring function of 0.1 and 1 % Low FPS?

flightlessmango:
In develop there is a new feature when you use the logging. It will display a window with some info of that logging session, including 1% lows. [example](https://cdn.discordapp.com/attachments/366023719199440908/717806164967948348/unknown.png)
Not exactly the same thing, but more than nothing

jazztickets:
So tried that in 0.4.1, and a blank window appears for about 6-7 seconds:

![log1](https://user-images.githubusercontent.com/425930/84428068-15ce7800-abe3-11ea-88c2-df29d34fda7a.png)

Then, for about 1 second the information appears, then disappears:
![log2](https://user-images.githubusercontent.com/425930/84428094-1ff07680-abe3-11ea-9941-792f555299b2.png)

I imagine that information should be appearing the whole time, correct?

Here's the config i'm using:
```
cpu_temp=1
gpu_temp=1
vram=1
ram=1
frame_timing=0
background_alpha=0
font_size=15
position=bottom-right
offset_y=-180
toggle_hud=F8
toggle_logging=F6
reload_cfg=F5
output_file=/var/log/mango/drg
```

flightlessmango:
@jazztickets 
Should be fixed in develop now

PR #232, opened 2020-06-18
add 0.1% lows to the brief benchmark overlay

kingdom5500:
i believe this would be the last change that issue #195 requested, but whether you close that issue is up to you of course. :)

i'm not sure why the existing code for getting 1% lows in the `calculate_benchmark_data` function uses a for loop to work it out. my solution here only uses one line to do the same thing for 0.1% lows. if there's any particular reason for the existing method being the way it is, i'll be happy to change my solution to be consistent with that. 

example output:
image

### DIFF
--- a/src/overlay.cpp
+++ b/src/overlay.cpp
@@ -789,6 +789,8 @@ void calculate_benchmark_data(){
       benchmark.total = sorted[i];
    }
    benchmark.oneP = benchmark.total;
+   // 0.1% min
+   benchmark.pointOneP = sorted[sorted.size() * 0.01];
 }
 
 void update_hud_info(struct swapchain_stats& sw_stats, struct overlay_params& params, uint32_t vendorID){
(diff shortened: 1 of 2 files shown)


flightlessmango:
Doesn't `benchHeight` need to be corrected for this?

kingdom5500:
oh of course, you're right. i'll adjust that now!

on a different note, this is the code used to calculate 1% lows:
```cpp
   for (size_t i = 0; i < sorted.size() * 0.1; i++){
      benchmark.total = sorted[i];
   }
```
why is `* 0.1` used here? wouldn't that be what you'd use for 10% lows? :eyes:

kingdom5500:
i've adjusted the `benchHeight` value, does this look fine to you? i want to make sure i get it right before i make the commit haha :)

image

flightlessmango:
Yes I think that was an oversight by me, doesn't your 0.1% also need an extra 0 in the calculation?
benchHeight looks good to me

Issue #229, opened 2020-06-17
Support more advanced logging


larskraemer:
I would like to know what you think about extending the logging feature a bit. Currently, the only values that can be logged are FPS and CPU/GPU load. It would be nice to let the user configure what is logged, including options like CPU/GPU temps, RAM, clocks, etc.

flightlessmango:
I don't see any problem with adding all the values to the log that are being monitored already. I think the question is what to do with them. Just logging the values is one thing, the question is what we do with them, are you envisioning a graph for every type of value?

larskraemer:
That would probably be a bit cluttered in the end. I think adding just the values the user requests to the log, and then graphing those values would be better. This would obviously require an additional row in the log file which specifies what values _are_ being logged.

mattlorimor:
@flightlessmango - re: this topic

You got a direct shout-out and some major praise from Linus and Anthony of Linus Tech Tips. Linus did mention, though, that he wished more values could be logged.

> Honestly, if it logged more information like CPU and GPU temperatures, we'd probably switch over to it instead of what we're using in Windows.

https://youtu.be/6T_-HMkgxt0?t=353

PR #234, opened 2020-06-18
Logging options

larskraemer:
Referencing #229. I added a few config options for what should be logged.
The code is a bit ugly; I didn't want to be too fancy because I don't really know the codebase.

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,7 +149,6 @@ A partial list of parameters are below. See the config file for a complete list.
 | `output_file`                      | Define name and location of the output file (Required for logging)                    |
 | `font_file`                        | Change default font (set location to .TTF/.OTF file )                                 |
 | `log_duration`                     | Set amount of time the logging will run for (in seconds)                              |
-| `log_fps` <br> `log_cpu_load` <br> `log_gpu_load` <br> `log_cpu_temp` <br> `log_gpu_temp` <br> `log_gpu_core_clock` <br> `log_gpu_mem_clock`  | Enable logging for different statistics |
 | `vsync`<br> `gl_vsync`             | Set vsync for OpenGL or Vulkan                                                        |
 | `media_player`                     | Show media player metadata                                                            |
 | `media_player_name`                | Set main media player DBus service name without the `org.mpris.MediaPlayer2` part, like `spotify`, `vlc`, `audacious` or `cantata`. Defaults to `spotify`. |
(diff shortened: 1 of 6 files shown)


flightlessmango:
I think everything can be logged regardless of if the param is enabled

larskraemer:
Should I just take out the config options then and make it log everything always?

flightlessmango:
Yes that sounds good

larskraemer:
Alright, apart from the messed up commit history, this should work fine.
Should I add more information to the log, like memory usage etc.?

jazztickets:
I'd like to see ram and vram added as well.

flightlessmango:
I think this is good, just needs to be resolved

jackun:
Mem usage needs a `loggingOn` check too

jackun:
@larskraemer can you rebase with `git rebase -i 58f0dc4040b83e4ba6f96cb9fa88d9966bbda064` or some such and drop unrelated commits and then rebase on `develop`? But at least git itself drops these as duplicates anyway when rebasing.

larskraemer:
Should be better now..

jackun:
Crap, at least amdgpu file handles aren't opened if they are initially disabled.
> 72.001,1,0,44,0,0,0,0,0,100129

larskraemer:
@jackun That is indeed a problem.. I changed the `init_gpu_stats` function to always try to initialize the GPU and disable the GPU overlay if it fails. In that case, we should just see '0' in the logs as before, which I think is sensible

flightlessmango:
I'm confused by f850ccfeaa756356bee98557cf9337486fd778d7, now it won't show current ram unless it's logging?

larskraemer:
It won't show RAM _unless_ it's logging? Does it at least not log the RAM anyway?

flightlessmango:
I misunderstood, the code is fine. We just need the same thing for a bunch of others that are disabled if the param is disabled

larskraemer:
I think I covered every param check for the things we are logging:

- `fps` is updated anyway
- `cpu_load` and `cpu_temp` are updated by `UpdateCPUData`, which has a check for `loggingOn`
- `gpu_load`, `gpu_temp`, `gpu_core_clock` and `gpu_mem_clock` are updated by `get*GpuInfo`, which has a check
- `ram_used` is updated by `update_meminfo`, which also has a check

jackun:
Did you have plans for overlay_params? Doesn't seem to get used currently.

jazztickets:
Can we get the .csv extension added to the log file while we're at it? It makes it easier for the folks who like to import the logs into LibreOffice calc or other programs. Although we'd want to wrap quotes around the values in the 2nd line of the log, where it displays the system info, just in case a comma appears in one of the values.

larskraemer:
Ah, I forgot to remove that, it was to allow the user to configure what gets logged. Since we are now logging everything, it's not needed anymore. I will remove it after rebasing onto develop... again

PR #236, opened 2020-06-19
implement configurable benchmark percentiles.

kingdom5500:
this is my first time writing more than like 20 lines of coherent C++ so i hope the code is alright haha. i'll be happy to make changes and take feedback though ^^

default config (`97+AVG+1+0.1`):
image

an example of quartiles (`75+50+25`):
image

empty config:
image

very excessive config to demonstrate height scaling:
image


kingdom5500:
i think PR #237 should be merged before this one, since this code could be changed to make use of that parser so that we keep all potential bugs in one place :D

flightlessmango:
#237 has been merged

kingdom5500:
sorry, i've been busy over the last few days. i messed up a little during the rebase since i've never done it before but i hope that it's not too much of an issue. i believe the code works properly, but it would be a good idea for somebody else to test it too in case i overlooked something :)

flightlessmango:
I'd like to merge this but I have been unable to verify that it works properly, I can't apply this as a patch for some reason

kingdom5500:
hmm, i'm not sure why that might be, or what you could do to solve it. can you not directly pull my branch and try to install mangohud from there? i suppose it's not ideal, but it could work if you haven't tried it yet.

flightlessmango:
I think I've managed to properly test this now, however it seems to have multiple errors when trying to build

flightlessmango:
This has been manually merged. Sorry it took so long time

PR #237, opened 2020-06-19
First implementation of custom float parsing

larskraemer:
This is an implementation of a floating-point parser. It is basically a state machine, and the code is written quite explicitly to allow for changes and additions to the state machine. It could probably be written more compactly and efficiently.
Some precision is lost depending on the input. If this is an issue for any application, that can probably be fixed.
Now for the good parts:

- This is thread-safe
- It does not depend on C-locale
- It will throw if the string is not a float
- It reports where the float ends

If there are no issues with the implementation, I would replace the other occurrences of any floating-point parsing with calls to this function before merging the commit.

### DIFF
--- a/src/string_utils.h
+++ b/src/string_utils.h
@@ -106,7 +106,7 @@ static bool try_stoull(unsigned long long& val, const std::string& str, std::siz
 
 static float parse_float(const std::string& s, std::size_t* float_len = nullptr){
     std::stringstream ss(s);
-    ss.imbue(std::locale("C"));
+    ss.imbue(std::locale::classic());
     float ret;
     ss >> ret;
     if(ss.fail()) throw std::invalid_argument("parse_float: Not a float");
--- a/src/string_utils.h
+++ b/src/string_utils.h
@@ -213,7 +213,7 @@ static float parse_float(const std::string& s, std::size_t* float_len = nullptr)
     if(float_len != nullptr) {
         *float_len = (it - s.begin());
     }
-    return result;
+    return result*sign;
 }
 
 #pragma GCC diagnostic pop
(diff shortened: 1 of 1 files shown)


jackun:
Is it any better?
I'd rather just 
```   float val = 0;
   std::stringstream ss(str);
   ss.imbue(std::locale("C"));
   ss >> val;```

**Comment by larskraemer** [comment]
While that works, you have no indication if you actually read a float. Also, you have to dump a whole string (which may or may not be large) into the string stream to use that method.
More important than the implementation of this function is, i believe, that it is implemented at all, so that it can be used.

**Comment by larskraemer** [comment]
It turns out, you can check for errors on a `stringstream`. Avoiding a single copy of a string is not enough reason to keep this mess around.

File: src/string_utils.h [hunk 1]
```diff
@@ -104,4 +104,18 @@ static bool try_stoull(unsigned long long& val, const std::string& str, std::siz
     return false;
 }
 
+static float parse_float(const std::string& s, std::size_t* float_len = nullptr){
+    std::stringstream ss(s);
+    ss.imbue(std::locale("C"));
```
jackun, line 109: Actually, `std::locale::classic()` could be preferred maybe

larskraemer: The two are equivalent, but you're right, this way we avoid a constructor call.

src/string_utils.h
```diff
@@ -104,4 +104,18 @@ static bool try_stoull(unsigned long long& val, const std::string& str, std::siz
...
 
+static float parse_float(const std::string& s, std::size_t* float_len = nullptr){
+    std::stringstream ss(s);
+    ss.imbue(std::locale::classic());
+    float ret;
+    ss >> ret;
+    if(ss.fail()) throw std::invalid_argument("parse_float: Not a float");
+    if(float_len != nullptr){
+        auto pos = ss.tellg();
+        if(ss.fail()) *float_len = s.size();
```
jackun, line 115: This fail() doesn't get checked if it throws before it.

larskraemer: The second fail check refers to `tellg`. If the `stringstream` is at the end of the string and you call `tellg` on it, it is supposed to return -1. It will also set the failbit.

PR #247, opened 2020-06-21
Move logging functionality into a class

larskraemer:
I moved most of the logging-related code into a class. This eliminates a bunch of global variables and also gives us a single point where we can change the code later.
There is now a global `unique_ptr<Logger>` which handles the logging. I wasn't quite sure where an appropriate entry-point would be, to construct the global, so I added checks in the functions that use it. If anyone who is more familiar with the codebase knows where that would be more appropriate to do, I will gladly change it.

### DIFF
--- a/src/overlay.cpp
+++ b/src/overlay.cpp
@@ -789,14 +789,15 @@ void check_keybinds(struct swapchain_stats& sw_stats, struct overlay_params& par
 #else
      pressed = false;
 #endif
-     if (pressed && (now - log_end > 11s)){
+     if (pressed && (now - logger->last_log_end() > 11s)) {
        last_f2_press = now;
-       if(logger->is_active()){
+       if (logger->is_active()) {
          logger->stop_logging();
-         std::thread(calculate_benchmark_data).detach();
        } else {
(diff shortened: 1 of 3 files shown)


flightlessmango:
Could probably put the output under `NDEBUG`

larskraemer:
Oh my bad, I was debugging something and forgot to take it out. I'll keep the one in the constructor for future troubleshooting and remove the rest.

larskraemer:
Alright, I cleaned up my code a bit. As far as I'm concerned, this is done for now.

src/overlay.cpp
```diff
@@ -1151,7 +1143,11 @@ void render_benchmark(swapchain_stats& data, struct overlay_params& params, ImVe
 
 void render_imgui(swapchain_stats& data, struct overlay_params& params, ImVec2& window_size, bool is_vulkan)
 {
+<<<<<<< HEAD
```
jackun, line 1146: Missed this

larskraemer: Thanks, my bad. Fixed now

PR #253, opened 2020-06-24
Use the std::chrono convenience classes for timing

larskraemer:
I created a `MesaClock` class to wrap the `os_time_get_nano` function and replaced most timing calculations with `chrono`s helper classes. This lets us write things like `elapsedTime >= chrono::seconds(12)`, or `elapsedTime >= 12s`, which is much easier to read than `elapsedTime/1000000 >= 12`.

@jackun I would like to use a `Clock::duration` for the values in the `overlay_params` struct, too, however it is declared as `extern C` (in fact, the whole header is). I'm not sure why that is, it doesn't seem like it is included in any C-files.

### DIFF
--- a/src/gl/inject_glx.cpp
+++ b/src/gl/inject_glx.cpp
@@ -127,10 +127,11 @@ EXPORT_C_(void) glXSwapBuffers(void* dpy, void* drawable) {
     do_imgui_swap(dpy, drawable);
     glx.SwapBuffers(dpy, drawable);
 
-    if (!is_blacklisted() && fps_limit_stats.targetFrameTime > 0){
-        fps_limit_stats.frameStart = os_time_get_nano();
+    using namespace std::chrono_literals;
+    if (!is_blacklisted() && fps_limit_stats.targetFrameTime > 0s){
+        fps_limit_stats.frameStart = Clock::now();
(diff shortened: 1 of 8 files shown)


jackun:
Don't remember but maybe std::chrono::high_resolution_clock etc already uses `clock_gettime`?

larskraemer:
It probably does. The only differences would probably be on Windows, if at all.
Using `high_resolution_clock` instead is as easy as changing the `using Clock=MesaClock;` line into `using Clock=std::chrono::high_resolution_clock;`
Frankly, I'd trust the standard library to provide a good high-resolution clock, so if you think we should use that instead, I'll change it.

jackun:
> I'd trust the standard library to provide a good high-resolution clock,

Hmm, probably needs some "research", but might as well use "MesaClock" for now, then at least you know what you get.

larskraemer:
According to Microsoft, `steady_clock` and `high_resolution_clock` use [`QueryPerformanceCounter`](https://docs.microsoft.com/en-us/cpp/standard-library/chrono?view=vs-2019), same as the implementation in `mesa/util/os_time.c`.
libcxx typedef's `high_resolution_clock` to `steady_clock`, which in turn uses `clock_gettime`

PR #257, opened 2020-06-25
Update cpu.cpp

AnorakMan:
fallback for k10temp method is passed the wrong path variable

### DIFF
--- a/src/cpu.cpp
+++ b/src/cpu.cpp
@@ -294,7 +294,7 @@ bool CPUStats::GetCpuFile() {
         if (name == "coretemp" && find_temp_input(path, input, "Package id 0")) {
             break;
         }
-        else if ((name == "zenpower" || name == "k10temp") && find_temp_input(path, input, "Tdie")) {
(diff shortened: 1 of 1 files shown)


jackun:
Wrong path in what way?
This also simplifies down to `else if (name == "zenpower" || name == "k10temp")`

AnorakMan:
On a system where the temperature file exists in "/sys/class/hwmon/hwmon2/temp1_input", but no Tdie file exists, then the loop exists with "path" set to hwmon0 instead of "hwmon2"

Then by the time the fallback method is run, it's looking in the wrong path.

jackun:
"Tdie" is not a file but contents of `tempN_label`. What's in `cat /sys/class/hwmon/hwmon2/temp1_label`?

AnorakMan:
I don't have a cat /sys/class/hwmon/hwmon2/temp1_label file

jackun:
`find_temp_input` might need to be separately actually. Try 

```
...
else if (name == "zenpower" || name == "k10temp") {
    find_temp_input(path, input, "Tdie");
    break;
}
...
```

AnorakMan:
Yeah, that ends up doing the same thing. My dodgy first attempt was just to force it to run the "find_temp_input" method before breaking. 
Thanks

Issue #254, opened 2020-06-24
"MANGOHUD_CONFIG=full" gives "Failed to get initial media player metadata."


rtentser:
Arch Linux, "Vampire: The Masquerade - Bloodlines". Mangohud works, but gives this output:

```
MANGOHUD: The name org.mpris.MediaPlayer2.spotify was not provided by any .service files
MANGOHUD: Failed to get initial media player metadata.
MANGOHUD: Connected to D-Bus as ":1.148".
MANGOHUD: The name org.mpris.MediaPlayer2.spotify was not provided by any .service files
MANGOHUD: Failed to get initial media player metadata.
MANGOHUD: The name org.mpris.MediaPlayer2.spotify was not provided by any .service files
MANGOHUD: Failed to get initial media player metadata.
MANGOHUD: The name org.mpris.MediaPlayer2.spotify was not provided by any .service files
MANGOHUD: Failed to get initial media player metadata.
MANGOHUD: XNVCtrl is using display :0
MANGOHUD: The name org.mpris.MediaPlayer2.spotify was not provided by any .service files
MANGOHUD: Failed to get initial media player metadata.
MANGOHUD: The name org.mpris.MediaPlayer2.spotify was not provided by any .service files
MANGOHUD: Failed to get initial media player metadata.
MANGOHUD: The name org.mpris.MediaPlayer2.spotify was not provided by any .service files
MANGOHUD: Failed to get initial media player metadata.
MANGOHUD: The name org.mpris.MediaPlayer2.spotify was not provided by any .service files
MANGOHUD: Failed to get initial media player metadata.
MANGOHUD: The name org.mpris.MediaPlayer2.spotify was not provided by any .service files
MANGOHUD: Failed to get initial media player metadata.
MANGOHUD: The name org.mpris.MediaPlayer2.spotify was not provided by any .service files
MANGOHUD: Failed to get initial media player metadata.
```

Strangely (or not), `MANGOHUD_CONFIG=cpu_temp,gpu_temp,core_load,gpu_core_clock,gpu_mem_clock,ram,vram` doesn't give the same result.

larskraemer:
Are you using `MANGOHUD_CONFIG=full` or similar? The warning appears because mangohud is trying to talk to a media player, but you don't appear to have one open.

rtentser:
> Are you using MANGOHUD_CONFIG=full or similar?

Yes.

> but you don't appear to have one open.

[...]

PR #264, opened 2020-06-30
Media Player autodetect

larskraemer:
#254
This is a draft, because I don't think we've reached a consensus what the media player functionality should look like. 
Right now, this PR removes the secondary media player display and lets MangoHud switch between media players.
`media_player_name` acts as a preferred media player. If that media player isn't running, any active one will be chosen.

### DIFF
--- a/src/dbus.cpp
+++ b/src/dbus.cpp
@@ -52,9 +52,14 @@ public:
     DBusMessageIter_wrap(DBusMessage* msg, libdbus_loader* loader)
     {
         m_DBus = loader;
-        m_DBus->message_iter_init(msg, &m_Iter);
-        m_resolved_iter = resolve_variants();
-        m_type = m_DBus->message_iter_get_arg_type(&m_resolved_iter);
+        if(msg){
+            m_DBus->message_iter_init(msg, &m_Iter);
+            m_resolved_iter = resolve_variants();
+            m_type = m_DBus->message_iter_get_arg_type(&m_resolved_iter);
+        }
+        else {
+            m_type = DBUS_TYPE_INVALID;
+        }
     }
 
     DBusMessageIter_wrap(DBusMessageIter iter, libdbus_loader* loader)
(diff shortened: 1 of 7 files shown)


src/dbus.cpp
```diff
@@ -657,37 +688,35 @@ void dbus_manager::start_thread()
...
+        if (!m_dbus_ldr.connection_read_write(m_dbus_conn, 0))
             return; // connection closed
 
-        msg = dbus.connection_pop_message(pmgr->m_dbus_conn);
+        msg = m_dbus_ldr.connection_pop_message(m_dbus_conn);
 
         // loop again if we haven't read a message
         if (nullptr == msg) {
-            std::this_thread::sleep_for(ms(10));
+            std::this_thread::yield();
```
jackun, line 710: dbus.connection_read_write() should actually have a timeout set then?

larskraemer: It should in this case, because the thread is doing nothing but dbus interaction. I tested it and it seemed to block the main thread. Until I figure out why, I left it like this

src/dbus.cpp
```diff
@@ -641,9 +640,43 @@ void dbus_manager::disconnect_from_signals()
...
+    m_dbus_ldr.error_init(&error);
+    if (nullptr == (dbus_reply = m_dbus_ldr.connection_send_with_reply_and_block(dbus_mgr.get_conn(), dbus_msg, DBUS_TIMEOUT, &error))) {
+        m_dbus_ldr.message_unref(dbus_msg);
+        std::cerr << "MANGOHUD: " << __func__ << ": "<< error.message << "\n";
+        m_dbus_ldr.error_free(&error);
+        return false;
+    }
+
+    auto iter = DBusMessageIter_wrap(dbus_reply, &m_dbus_ldr);
+    if(not iter.is_array()) return false;
```
jackun, line 668: Leaking dbus_msg etc, I think

larskraemer: You're right. I'm currently looking into writing a (much smaller) wrapper for DBusMessage that cleans up after itself, but for the time being, I'll fix this.
Good documentation on the low-level C API is surprisingly hard to find. I'll have to figure out which messages need to be unref'd and which don't.

src/dbus.cpp
```diff
@@ -610,6 +519,94 @@ dbus_manager::~dbus_manager()
...
+        }
+    }
+    return DBUS_HANDLER_RESULT_NOT_YET_HANDLED;
+}
+
+bool dbus_manager::handle_properties_changed(DBusMessage* msg, const char* sender) {
+    std::string source;
+
+    metadata meta;
+    parse_mpris_properties(m_dbus_ldr, msg, source, meta);
```
jackun, line 541: Properly behaving players may just send "PlaybackStatus" so this just clears out metadata.

larskraemer: Okay, didn't know that. Should I just copy the old metadata and have `parse_mpris_properties` overwrite it?

jackun: Maybe. Depends how it meshes with the m_active_player idea

larskraemer: I think we should be fine if we clear the metadata completely if we change the active player.
Do you know a player which doesn't send full status updates, so I can test that situation?

jackun: Most not named "Spotify" but i'm using Audacious and VLC.

larskraemer: Not sure how I missed that, then, but yes, VLC breaks it. I'm almost finished with the message wrapper, I'll commit a fix after that one.

larskraemer:
@jackun Please ignore the debug output for now, I have a feeling I'm going to need it. :D I'll remove it when the PR is ready.
I'm going to test this with Audacious, but if you still get some wrong behaviour, please let me know

src/dbus.cpp
```diff
@@ -582,14 +525,45 @@ bool dbus_manager::init(const std::string& dest)
...
+        return true;
+    }
+
+    // Else, use any player that is currently playing..
+    for(const auto& entry : m_name_owners) {
+        const auto& name = std::get<0>(entry);
+        metadata meta;
+        get_media_player_metadata(meta, name);
+        if(meta.playing) {
+            m_active_player = name;
```
jackun, line 555: Maybe even `main_metadata.meta = meta;` here, for when audacious etc is paused at startup

larskraemer: Just solved this by passing in a pointer to a `metadata` and storing the updated values there.

larskraemer:
@jackun 
I changed the logic to use a fallback player only if no player is requested, as you suggested on discord.
There is a small issue; If two players are playing at the same time and no player is requested, which one gets shown depends on the implementation of `std::unordered_map`.
If you're fine with that, I'll clean everything up and make it ready for merging.

PR #273, opened 2020-07-06
simplify cross arch preloading

### LINKED ISSUE #218: Easier 32bit support
I just noticed that the mangohud bin script is a bit complicated when it comes to 32bit support. I would like to see it more generic, since that would also includes other architectures like arm etc. [In gamemode, we do it like this](https://github.com/FeralInteractive/gamemode#requesting-gamemode):
```sh
LD_PRELOAD="$LD_PRELOAD:/usr/\$LIB/libgamemodeauto.so.0"
```
Or more precisely, that's the end result for a normal setup.

In reality the script looks like this:
```
GAMEMODEAUTO="@GAMEMODERUN_PREFIX@libgamemodeauto.so.0"

LD_PRELOAD="${GAMEMODEAUTO}${LD_PRELOAD:+:$LD_PRELOAD}"

exec $GAMEMODERUNEXEC env LD_PRELOAD="${LD_PRELOAD}" "$@"
```
where `@GAMEMODERUN_PREFIX@` is configured via Meson, basically it's `prefix+'\$LIB/'` (note that the backslash is important here).

This removes the need of checking the target machine in Meson as well.

/cc @flightlessmango @jackun

stephanlachnit:
Closes #218.

Basically, when passing `$LIB` to `ld` it will be replaced with the systems libdir (just like with in Meson, without prefix). Because we need to pass `$LIB`, the backslash before is required to ignore the `$`.
With this, any cpu-arch specific stuff can be removed from Meson and the launch script.

/cc @jackun @flightlessmango


stephanlachnit:
Actually I might want to check if the same thing can be done for https://github.com/flightlessmango/MangoHud/blob/develop/src/mangohud.json.in, i. e. setting `"library_path": "mangohud/libMangoHud.so",`.

jackun:
Something in `ld-preload` branch. Need some annoying distro-agnostic tweaks for build.sh prolly

stephanlachnit:
> Something in ld-preload branch. Need some annoying distro-agnostic tweaks for build.sh prolly

Oh I didn't see that branch. Anyway, why don't you just use `sudo meson install -C builddir` instead of doing it for every file on it's own? Meson is preconfigured with the correct library paths (i. e. no need to check the distro for that), so it wouldn't require any tweaks.

jackun:
Yeah, probably works too. Just to keep it the same with prebuilt tar so less guessing where the hell it actually installs

stephanlachnit:
> Yeah, probably works too. Just to keep it the same with prebuilt tar so less guessing where the hell it actually installs

Hm, Meson does output every file it installs, and it also has an uninstall option.
I'll take a look and maybe simplify build.sh in such a way that it sill keeps the prebuild tar.

stephanlachnit:
Ok from my tests, it will now work without the need of any cross arch preloading. It _probably_ breaks `build.sh` in some ways (i. e. all the `.x86` files are gone), but overall it shouldn't break to hard.
Can't really test the script since I'm not on an FHS 3.0 based system anyway.

jackun:
Does [ld-preload-2](https://github.com/flightlessmango/MangoHud/tree/ld-preload-2) still work? Looked like `append_libdir_mangohud` isn't used currently so renamed/reused it for `build.sh` compatibility.

stephanlachnit:
> Does [ld-preload-2](https://github.com/flightlessmango/MangoHud/tree/ld-preload-2) still work? Looked like `append_libdir_mangohud` isn't used currently so renamed/reused it for `build.sh` compatibility.

Just commented on https://github.com/flightlessmango/MangoHud/commit/1436cc6acfdab200781a84e0d5fdc52fa4b43646 :)

jackun:
Major Pita, U.S. Air Force.
Wait, `append_libdir_mangohud` was tru by default, debian packaging used it?

stephanlachnit:
> Wait, `append_libdir_mangohud` was tru by default, debian packaging used it?

I did, yes. Kinda unintentionally at the beginning, but when I took a deeper look it was exactly what I needed^^

stephanlachnit:
Btw I think ld-preload-2 looks good now :+1

stephanlachnit:
@jackun anything missing to get this merged into develop?

jackun:
Idk about build.sh. Probably just try to guess a symlink for $LIB.

jackun:
Should be https://github.com/flightlessmango/MangoHud/commit/d712d35aaf5d548f23cc1e1596646cbafc4004a3 so prolly can be closed here.

PR #286, opened 2020-07-18
Add wine version and wine colors to overlay

gort818:
Allows the ability to show current version running, and to change the color. Not functioning with proton versions yet.


rawfoxDE:
Good job man !

jackun:
Does using `get_exe_path()` and getting `wine` path from `wine[64]-preloader`  aka /proc/self/exe work?

gort818:
I will give that a shot.
Sure no problem.

src/overlay.cpp
```diff
@@ -1301,6 +1325,15 @@ void render_imgui(swapchain_stats& data, struct overlay_params& params, ImVec2&
          ImGui::PopFont();
       }
 
+       if (params.enabled[OVERLAY_PARAM_ENABLED_wine]){
+           ImGui::TextColored(ImGui::ColorConvertU32ToFloat4(params.wine_color), "%s","WINE");
+           ImGui::PushFont(data.font1);
+           //ImGui::Dummy(ImVec2(0, 8.0f));
+           auto wine_color = ImGui::ColorConvertU32ToFloat4(params.wine_color);
```
flightlessmango, line 1332: No need to convert the `wine_color` twice

gort818:
Well I believe I screwed up that rebase. Should I just work off the develop branch then submit another PR to merge into develop?

jackun:
:D Just `git rebase -i 5ce6bcb79cb0dd1de41c79fcfa8a410d0dde645e` and drop the other unrelated commits, probably.

gort818:
Well I just made it worse :( so sorry

gort818:
Well that is a little better, not sure how that readme change ended up there..oops

jackun:
"Merge branch 'develop' into wine2" should be probably squashed into "Add wine and wine colors to overlay"

src/overlay.cpp
```diff
@@ -1341,6 +1365,13 @@ void render_imgui(swapchain_stats& data, struct overlay_params& params, ImVec2&
          ImGui::PopFont();
       }
 
+       if (params.enabled[OVERLAY_PARAM_ENABLED_wine]){
+           ImGui::TextColored(ImGui::ColorConvertU32ToFloat4(params.wine_color), "%s","WINE");
+           ImGui::PushFont(data.font1);
+           ImGui::TextColored(wine_color, "%s", wineVersion.c_str());
```
flightlessmango, line 1371: `wine_color` doesn't exist now, the first conversion is not stored

gort818:
Ok, I think I got it squared away. Thanks for the help!

flightlessmango:
It would be good if this doesn't show when it's not wine

gort818:
At the moment it will display "N/A" if wine is enabled in the config at not runnig, you want it to not display at all if wine is not being used?

jackun:
It is probably better to use `get_exe_name()` and derive `wine`'s path from it. There could be bajillion other wineservers running.

gort818:
Yeah was testing that out this past weekend, I will get something together.

flightlessmango:
> At the moment it will display "N/A" if wine is enabled in the config at not runnig, you want it to not display at all if wine is not being used?

Yes, I think that would be better

gort818:
Awesome, thanks for clarifying! I will get on it.

gort818:
Working on getting Proton version to display as well, should have it ready shortly.

gort818:
When the wine option is enable here is how it looks.

MangoHud showing wine version running Star Citizen
wine-mango
MangoHud not showing wine version running vkcube
mango-native
MangoHud showing Proton version running Doom Eternal 
mango-proton

src/overlay.cpp
```diff
@@ -689,6 +689,52 @@ void init_system_info(){
       trim(gpu);
       driver = exec("glxinfo | grep 'OpenGL version' | sed 's/^.*: //' | cut -d' ' --output-delimiter=$'\n' -f1- | grep -v '(' | grep -v ')' | tr '\n' ' ' | cut -c 1-");
       trim(driver);
+
+// Get WINE version
+
+      wineProcess = get_exe_path();
+      trim(wineProcess);
```
jackun, line 696: Shouldn't be needed probably. Any whitespace is actually in the file's name then.

jackun:
Too much `sed` for my liking :D Maybe something like this: 
[wine-version.diff.txt](https://github.com/flightlessmango/MangoHud/files/5074579/wine-version.diff.txt)

src/overlay.cpp
```diff
@@ -694,24 +694,40 @@ void init_system_info(){
...
+      if (preloader == "wine-preloader" || preloader == "wine64-preloader") {
+         // Check if using Proton
+         if (wineProcess.find(searchProton) != std::string::npos) {
+            stringstream protonPath;
+            protonPath << "echo " << wineProcess <<" | rev | cut -d '/' -f3- | rev | sed 's/.*/\"&\"/'";
+            string protonVersion;
+            protonVersion = exec(protonPath.str());
+            trim(protonVersion);
+            stringstream command;
+            command <<"< " <<protonVersion + "/version" <<" cut -d' ' -f2";
```
flightlessmango, line 709: Thinking something like this to make it prettier and conform more to how it's displayed in Steam
```
command <<"< " <<protonVersion + "/version" <<" cut -d' ' -f2 | sed -e 's|proton-||' ";
wineVersion = exec(command.str());
wineVersion = "Proton " + wineVersion;
```

gort818:
I might have gone a little overboard with sed:). Awesome I will take a look.

src/overlay.cpp
gort818: Yes that is much nicer.. also was thinking what if we replace WINE with PROTON and then  display the proton version underneath?
`` ImGui::TextColored(wine_color, "%s", "PROTON");``

flightlessmango: I'm not sure the WINE/PROTON text is required, the version should always contain proton or wine in it already

gort818:
MangoHud full with custom proton
mangonew
MangoHud full with default proton
mangonew1

flightlessmango:
Everything looks a-ok to me :ok_hand:

gort818:
Great to hear!

jackun:
Should be in https://github.com/flightlessmango/MangoHud/commit/632f1e7000a68602ce3410ab987a130953b5b7d7

PR #294, opened 2020-07-25
man page for mangohud

stephanlachnit:
I found a Debian Developer who is willing to upload MangoHud, but he wants a man page for `mangohud`. So I wrote one up, and it would be nice if it would be included in the next release so I don't need to add it manually.

If you want to view how it actually looks, run `man -l data/mangohud.1`.

/cc @jackun

### DIFF
--- a/data/mangohud.1
+++ b/data/mangohud.1
@@ -1,0 +1,53 @@
+.\" Manpage for mangohud.
+.TH mangohud 1 "" "" "mangohud"
+
+.SH NAME
+mangohud \- enable MangoHud on any application
+
+.SH SYNOPSIS
+\fBmangohud\fR COMMAND
+
+.SH DESCRIPTION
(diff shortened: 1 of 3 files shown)


data/mangohud.1
```diff
@@ -0,0 +1,53 @@
...
+.br
+Vulkan: \fBMANGOHUD=1 vkcube\fR
+.br
+Steam: set your launch option to \fBmangohud %command%\fR
+.br
+Lutris: add \fBmangohud\fR to the Command prefix setting
+.br
+OpenGL with dlsym: \fBMANGOHUD_DLSYM=1 mangohud glxgears\fR
+.br
+Custom config options: \fBMANGOHUD_CONFIG="gpu_stats=off,font_size=12" mangohud glxgears\fR
```
jackun, line 48: > gpu_stats=off

Technically works but currently it expects an int, so `gpu_stats=0`

stephanlachnit: Oh right, thanks!

PR #330, opened 2020-08-26
Toggle betwen fps_limit and unlimited or alternate fps limit via hotkey

ZakMcKrack3n:
Implemented this to better capture game footage (specifically racing game replays) where I needed stable fps , but I also am not able to close the game itself , so this enables the fps limit on/off functionality or alternate between 2 frame rates e.g. palying the game with `fps_limit` , toggle to `fps_limit_alt` to capture replay etc.

So far no documentation changes and or readme updates because naming/key combo is most probably not final.

### DIFF
--- a/src/keybinds.h
+++ b/src/keybinds.h
@@ -9,7 +9,7 @@
 typedef unsigned long KeySym;
 #endif
 
-Clock::time_point last_f2_press, last_f12_press, reload_cfg_press, last_upload_press;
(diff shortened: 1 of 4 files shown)


flightlessmango:
I think a better implementation would be to have fps_limit accept multiple values with a delimiter, like `benchmark_percentiles` for example. So every time you press the button you would cycle through the values

ZakMcKrack3n:
Ok , will try implement this , so e.g. 0,30,60 would be unlimited, 30 fps, 60 fps etc.

PR #337, opened 2020-08-30
Toggle between multiple fps limits

ZakMcKrack3n:
Refactored fps toggle to use multiple values for `fps_limit` , existing config files will still set fps limit but toggle will be inert.

New format: `0, 60 ,30` would toggle between unlimited, 60fps, 30fps

Caveats: Multiple values of the same value will cause switching between first occurrence of the value:
`0,30,60,48,60,72` would cycle from unlimited -> 30 -> 60 -> 48 -> 60 -> 48 .... 60 -> 48 forever
I am currently not versed enough in C to introduce a variable and to remember the index position to support this.


jackun:
Probably should rebase on `develop`, not merge. I cleaned it up a bit in `pr-337` 1bf35d73ea2fdfee71598f3121fc71c84a5a80a6 or feel free to try yourself :D

ZakMcKrack3n:
Ok will try to do that.

jackun:
FYI, merged in https://github.com/flightlessmango/MangoHud/commit/73e6e3130188806a4590a7f5f65e1bf2908b087f
Seems to work.

ZakMcKrack3n:
Yes , just started to rebase and saw its already in develop , thanks for your support !

PR #342, opened 2020-09-08
Add output_folder config option to MangoHud.conf

krathalan:
I was trying to find the default location where logs were placed, and I checked on your discord and you mentioned on 9/04 that the `output_folder` param must be set. I tried it and it's working now, but I figured it should be added to the default/sample config provided.

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,10 +198,13 @@ All vulkan vsync options might not be supported on your device, you can check wh
 
 ## MangoHud FPS logging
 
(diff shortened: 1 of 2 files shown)


flightlessmango:
`output_file` is deprecated and should be completely removed. readme should also be updated to mention `output_folder` instead in the logging section as well

krathalan:
Like that?

flightlessmango:
Looks good :ok_hand:

PR #356, opened 2020-09-28
fix typo in man page

stephanlachnit:
@jackun


jackun:
Thanks, typoglycaemia.

PR #376, opened 2020-11-02
Fix for FPS limit breaking

### LINKED ISSUE #127: Game breaks thru framerate limit
So I have 48 fps frame rate limit in my config file and it works as intended in Dark Souls Remastered and Dark Souls SotFS. Today I noticed that when I play Tomb Raider 2013 with proton 4.11.13 the cap works, but at some point it stops to work and it turn on again only if I go to main menu or die in the game.

Also "mangohud %command%" doesn't seem to work with OpenGL 
I am running Pop!_OS 19.10

Thanks

flgx16:
Checks if sleep time is higher than 0 which should be false if FPS is lower than FPS limit
With original logic if thread oversleeps longer than normally FPS limit would "break" until `stats.sleepTime > stats.frameOverhead` was true again

Fixes #127

However, this logic may cause FPS to slightly drop under set FPS limit (which is what the original frame overhead logic is supposed to prevent I assume) when FPS is only slightly higher than FPS limit

### DIFF
--- a/src/vulkan.cpp
+++ b/src/vulkan.cpp
@@ -2425,7 +2425,7 @@ void FpsLimiter(struct fps_limit& stats){
       auto adjustedSleep = stats.sleepTime - stats.frameOverhead;
       this_thread::sleep_for(adjustedSleep);
       stats.frameOverhead = ((Clock::now() - stats.frameStart) - adjustedSleep);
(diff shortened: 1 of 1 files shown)


flgx16:
Force-pushed to use @jackun's solution for this. That should fix it (at least mostly) without removing part of frame overhead logic.

Issue #64, opened 2020-03-14
New feature: Changing order of options


treier57lol:
Don't know if this has already been requested nor the best way to implement this. 

Probably the order of the options could work as an implementation. For configs that would be from top to bottom (for obvious reasons) and for the MANGOHUD_CONFIG variable left to right. This might not be the best option but I don't know, just throwing it out there.

flightlessmango:
This is now added into develop see #379

PR #379, opened 2020-11-03
Sorting hud elements

flightlessmango:
#64 This makes it possible to set the order the hud elements appear in. It sorts after the order they appear in the config file.
This also makes it possible to add multiple of the same hud element (whatever that would be good for) and currently everything needs to be manually enabled or the hud is empty.

TODO: add media_player

### DIFF
--- a/src/overlay.cpp
+++ b/src/overlay.cpp
@@ -14,6 +14,7 @@
 struct benchmark_stats benchmark;
 struct fps_limit fps_limit_stats {};
 ImVec2 real_font_size;
+std::vector<logData> graph_data;
 
 void update_hw_info(struct swapchain_stats& sw_stats, struct overlay_params& params, uint32_t vendorID)
 {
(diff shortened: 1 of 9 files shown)


src/overlay.cpp
```diff
@@ -55,15 +56,18 @@ void update_hw_info(struct swapchain_stats& sw_stats, struct overlay_params& par
 
    currentLogData.cpu_load = cpuStats.GetCPUDataTotal().percent;
    currentLogData.cpu_temp = cpuStats.GetCPUDataTotal().temp;
-
+   // Save data for graphs
+   if (graph_data.size() > 50)
+      graph_data.erase(graph_data.begin());
```
jackun, line 61: Could just graph_data.clear()

flightlessmango:
I'm pretty happy with this now, anything else before merge?

PR #380, opened 2020-11-03
Gradual colors for gpu & cpu load

gort818:
Gradually change colors for cpu and gpu load using  linear interpolation.


jackun:
@gort818 Can you instead rebase it on `develop` to get rid of ac202181eaf8467a83786ec99d4995aadaed8945? Seemed you'd only need to remove `render_mango` function in conflicting commit.
Squash and merge seems to work too though.

PR #388, opened 2020-11-11
Add cpu_power

flgx16:
Calculates CPU draw in watts using voltage & current sensors.

Currently supports only `k10temp`.

### DIFF
--- a/src/cpu.cpp
+++ b/src/cpu.cpp
@@ -240,6 +240,113 @@ bool CPUStats::UpdateCpuTemp() {
     return ret;
 }
 
+static bool get_cpu_power_k10temp(CPUPowerData* cpuPowerData, int& power) {
+    CPUPowerData_k10temp* powerData_k10temp = (CPUPowerData_k10temp*)cpuPowerData;
+
+    if (!powerData_k10temp->coreVoltageFile || !powerData_k10temp->coreCurrentFile || !powerData_k10temp->socVoltageFile || !powerData_k10temp->socCurrentFile)
+        return false;
+
+    rewind(powerData_k10temp->coreVoltageFile);
+    rewind(powerData_k10temp->coreCurrentFile);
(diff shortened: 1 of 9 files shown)


flgx16:
Cleaned it up, supports `k10temp` and `zenpower`.

flgx16:
Added Intel RAPL support and removed `path.clear()` because I don't think it's needed there. Not sure but I think that RAPL energy reading is gonna be restricted soon because of a new CPU vulnerability (PLATYPUS).

src/cpu.h
```diff
@@ -70,6 +139,7 @@ class CPUStats
    bool m_updatedCPUs = false; // TODO use caching or just update?
    bool m_inited = false;
    FILE *m_cpuTempFile = nullptr;
+   CPUPowerData* m_cpuPowerData = nullptr;
```
jackun, line 142: std::unique_ptr<CPUPowerData> m_cpuPowerData;

src/cpu.cpp
```diff
@@ -240,6 +242,113 @@ bool CPUStats::UpdateCpuTemp() {
...
+
+bool CPUStats::UpdateCpuPower() {
+    if(!m_cpuPowerData)
+        return false;
+
+    int power = 0;
+
+    switch(m_cpuPowerData->source) {
+        case CPU_POWER_K10TEMP:
+            if (!get_cpu_power_k10temp(m_cpuPowerData, power)) return false;
```
jackun, line 335: `m_cpuPowerData.get()` and others

jackun, line 335: `m_cpuPowerData.get()`

longusnickus:
Manjaro updated to Kernel 5.10. now i got max 30-40Watt at full load. It should be around 150Watt at full load on my 3700X

flgx16:
Can you create a new issue for that please?

PR #395, opened 2020-11-15
Add param to enable/disable fps milliseconds text

gort818:
set `show_ms = 0` in the config file to remove the milliseconds for the fps param

Set enabled by default

mango_ms


flightlessmango:
I think it would be better to have a bool for showing `ms` and have it set to true by default

gort818:
Yeah good idea, doing now

flightlessmango:
It would be more descriptive and in line to call it just `frametime` imo

flightlessmango:
Documenting it in readme and config file would be A+

Sporif:
> It would be more descriptive and in line to call it just `frametime` imo

Would be a bit confusing considering there's already `frame_timing` for the frametime graph

gort818:
Yeah `frametime` makes more sense.. but it is very close to `frame_timing`  and maybe should probably be `frametime_graph`

but for now I will make it frametime and add it the readme and config file

PR #398, opened 2020-11-18
Allow users to add a new item to the blacklist from config file

gort818:


src/vulkan.cpp
```diff
@@ -2079,8 +2079,17 @@ static VkResult overlay_CreateInstance(
                              &instance_data->vtable);
    instance_data_map_physical_devices(instance_data, true);
 
+   parse_overlay_config(&instance_data->params, getenv("MANGOHUD_CONFIG"));
+
+   //check for blacklist item in the config file
+   std::stringstream ss(instance_data->params.blacklist);
+   std::string token;
+   while (std::getline(ss, token, ',')) {
```
jackun, line 2087: Then probably make this `parse_str_tokenized` (or something, i can't name things :D ) parsed into a vector, for other APIs.

src/vulkan.cpp
```diff
@@ -2079,8 +2079,17 @@ static VkResult overlay_CreateInstance(
...
+      trim(token);
+      add_blacklist(token);
```
jackun, line 2089: Also vkCreateInstance can get called more then once so blacklist just gonna keep growing

src/vulkan.cpp
gort818: Ok now we have parse_str_tokenize .. I am not any better :)

src/vulkan.cpp
gort818: Hmm I am not seeing any duplicates in the blacklist, if that is what you meant.

gort818: Ok implemented a check if item exits in blacklist

README.md
```diff
@@ -188,7 +188,7 @@ Parameters that are enabled by default have to be explicitly disabled. These (cu
 | `cellpadding_y`                    | Set the vertical cellpadding, default is `-0.085` |
 | `frametime`                        | Display frametime next to fps text                                                    |
 | `table_columns`                    | Set the number of table columns for ImGui, defaults to 3                              |
-| `blacklist`                        | Add a program to the blacklist. e.g `blacklist=vkcube,WatchDogs2.exe`                 |
+| `blacklist`                        | Add a program to the blacklist. e.g `blacklist=vkcube + WatchDogs2.exe`                 |
```
flightlessmango, line 191: shouldn't be spaces between the blacklist names and the `+`

gort818: Because of aesthetics? or functionality?  `parse_str_tokenize` is calling trim() on the string

flightlessmango: I'm assuming it would not work well with the env var

gort818: Ahh ok yep makes sense

PR #423, opened 2020-12-08
fix: cut --output-delimiter

AlexxandreFS:
I have a problem with the command
cut --- output-delimiter
The busysbox does not support the "output-delimiter" parameter and I get an error message on the terminal.
print_2020-12-08_05 06 41


gort818:
Works here on gnu
maybe let's make it more concise 
Does this work for you? 
`driver = exec (glxinfo | grep 'OpenGL version' |sed 's/^.*: //' | sed 's/([^()]*)//g' |  tr -s ' ')`

AlexxandreFS:
@gort818 Confirmed, it works without any problem.
I think the way you did it got better.

PR #429, opened 2020-12-12
Add include_cfg

flgx16:
Allows including another configs (subconfigs) from a config.

Example:
`$HOME/.config/MangoHud/MangoHud.conf`:
```
cpu_stats
fps
fps_limit=80
```

`$HOME/.config/MangoHud/vkcube.conf`:
```
include_cfg=MangoHud
fps_limit=100
```

`vkcube` with MangoHud should now be limited to 100 FPS.

This PR doesn't implement watching all those configs for changes to automatically reload config(s).

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -3,21 +3,27 @@
 #include <iostream>
 #include <string.h>
 #include <thread>
-#include <unordered_map>
 #include <string>
 #include "config.h"
 #include "file_utils.h"
 #include "string_utils.h"
 #include "hud_elements.h"
 
-static void parseConfigLine(std::string line, std::unordered_map<std::string, std::string>& options) {
+std::vector<std::string> used_configs;
+static const char *mangohud_dir = "/MangoHud/";
+
+void enumerate_config_files(std::vector<std::string>& paths, std::string configName, bool subconfig);
+bool parseConfigFile(overlay_params& params, std::string path);
+
+static void parseConfigLine(std::string line, overlay_params& params) {
     std::string param, value;
 
     if (line.find("#") != std::string::npos)
         line = line.erase(line.find("#"), std::string::npos);
 
     size_t equal = line.find("=");
-    if (equal == std::string::npos)
+    bool no_value = equal == std::string::npos;
+    if (no_value)
         value = "1";
     else
         value = line.substr(equal+1);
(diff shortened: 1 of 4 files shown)


baryluk:
I would say this is a bad idea IMHO.

Instead of adding support for include_cfg (what about recursion, or error), simply read MangoHud.conf first, then game specific one, overriding any set values. Easy.

flgx16:
> I would say this is a bad idea IMHO.
> 
> Instead of adding support for include_cfg (what about recursion, or error), simply read MangoHud.conf first, then game-specific one, overriding any set values. Easy.

This is a little bit more complex solution than reading MangoHud.conf and then game specific config, which some people may see more useful than reading MangoHud.conf and then game-specific config.
Not sure what you mean by "what about recursion, or error", one config can not be used more than one time so I think that should be fine. https://github.com/flightlessmango/MangoHud/blob/dc11ee1b84220132745ae63506e368678f73a176/src/config.cpp#L85-L90

baryluk:
The code looks good to me. I don't have too big objections. I just don't have the use personally for it too much.

Is there some deeper motivation behind this support? To make it easier to configure from GOverlay maybe?

src/config.cpp
```diff
@@ -7,16 +7,24 @@
 #include "file_utils.h"
 #include "string_utils.h"
 #include "hud_elements.h"
+
 std::string program_name;
+std::vector<std::string> used_configs;
+static const char *mangohud_dir = "/MangoHud/";
+static std::string env_config = get_config_dir();
```
baryluk, line 14: That is a bit dangerous.

I don't think we should be using static non-trivial initializers.

flgx16:
> The code looks good to me. I don't have too big objections. I just don't have the use personally for it too much.
> 
> Is there some deeper motivation behind this support? To make it easier to configure from GOverlay maybe?

The motivation behind this was to not have to update configs which mostly had same parameters just with slight changes (such as fps limit) just to add a new parameter.
Now that you mentioned GOverlay, this PR may also be useful for GOverlay but probably not as much as for people editing configs manually.

baryluk:
> > The code looks good to me. I don't have too big objections. I just don't have the use personally for it too much.
> > Is there some deeper motivation behind this support? To make it easier to configure from GOverlay maybe?
> 
> The motivation behind this was to not have to update configs which mostly had same parameters just with slight changes (such as fps limit) just to add a new parameter.
> Now that you mentioned GOverlay, this PR may also be useful for GOverlay but probably not as much as for people editing configs manually.

Thanks for the clarifications.

src/config.cpp
flgx16: Does it look fine now?

flgx16:
Rebased.

PR #430, opened 2020-12-12
Allow using whitespace as divider between param and value

flgx16:


baryluk:
I don't think this is a good idea.

A better way would be instead to find just `'='`, then do a trim (aka `strip`) on the value.

I don't think it even solves any real problem.

PR #432, opened 2020-12-13
Improve code for program name extraction

baryluk:
This cleans up the code, as well fixes the bug of not setting
program_name if the MANGOHUD_CONFIG env is specified.

Instead of global variable, that could be not-initialized
use a function to get a program name in logging.

While at it, revamp code and separate things into own functions,
and return by value.


baryluk:
Oops. Please don't merge. I noticed commit missing from this PR.

baryluk:
Merge issues. Sorry. I will send it later.

PR #433, opened 2020-12-13
Improve code for program name extraction

baryluk:
This cleans up the code, as well fixes the bug of not setting
program_name if the MANGOHUD_CONFIG env is specified.
Otherwise, it would make program_name == "",
and log to files without a prefix.

Instead of global variable, that could be not-initialized
use a function to get a program name in logging.

While at it, revamp code and separate things into own functions,
and return by value, and make them easier to use / share code.

### DIFF
--- a/src/blacklist.cpp
+++ b/src/blacklist.cpp
@@ -7,19 +7,13 @@
 #include "file_utils.h"
 
 static std::string get_proc_name() {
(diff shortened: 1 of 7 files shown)


baryluk:
@flightlessmango Could you assign somebody to do a review of this?

It is blocking me from doing some other PRs. And I tried to make this one small and clean.

PR #434, opened 2020-12-13
Use raw clocks if available

Joshua-Ashton:
Avoids NTP synchronization having impact on frametime pacing.

### DIFF
--- a/src/mesa/util/os_time.c
+++ b/src/mesa/util/os_time.c
@@ -54,6 +54,13 @@
 #  error Unsupported OS
 #endif
 
+#if defined(CLOCK_MONOTONIC_RAW) /* linux */
+#define MANGOHUD_CLOCKTYPE CLOCK_MONOTONIC_RAW
+#elif defined(CLOCK_MONOTONIC_FAST) /* freebsd */
+#define MANGOHUD_CLOCKTYPE CLOCK_MONOTONIC_FAST
+#else /* fallback */
+#define MANGOHUD_CLOCKTYPE CLOCK_MONOTONIC
+#endif
 
 int64_t
 os_time_get_nano(void)
@@ -61,7 +68,7 @@ os_time_get_nano(void)
 #if DETECT_OS_LINUX || DETECT_OS_BSD
 
    struct timespec tv;
-   clock_gettime(CLOCK_MONOTONIC, &tv);
+   clock_gettime(MANGOHUD_CLOCKTYPE, &tv);
    return tv.tv_nsec + tv.tv_sec*INT64_C(1000000000);
 
 #elif DETECT_OS_UNIX
@@ -102,7 +109,7 @@ os_time_sleep(int64_t usecs)
    struct timespec time;
    time.tv_sec = usecs / 1000000;
    time.tv_nsec = (usecs % 1000000) * 1000;
-   while (clock_nanosleep(CLOCK_MONOTONIC, 0, &time, &time) == EINTR);
+   while (clock_nanosleep(MANGOHUD_CLOCKTYPE, 0, &time, &time) == EINTR);
 
 #elif DETECT_OS_UNIX
    usleep(usecs);
--- a/src/overlay_params.cpp
+++ b/src/overlay_params.cpp
@@ -132,7 +132,7 @@ parse_string_to_keysym_vec(const char *str)
 static uint32_t
 parse_fps_sampling_period(const char *str)
 {
-   return strtol(str, NULL, 0) * 1000;
+   return strtol(str, NULL, 0) * 1000000; /* ms to ns */
 }
 
 static std::vector<std::uint32_t>
@@ -538,7 +538,7 @@ parse_overlay_config(struct overlay_params *params,
    params->enabled[OVERLAY_PARAM_ENABLED_cpu_load_change] = false;
    params->enabled[OVERLAY_PARAM_ENABLED_legacy_layout] = true;
    params->enabled[OVERLAY_PARAM_ENABLED_frametime] = true;
-   params->fps_sampling_period = 500000; /* 500ms */
+   params->fps_sampling_period = 500000000; /* 500ms */
    params->width = 0;
    params->height = 140;
    params->control = -1;
--- a/src/hud_elements.cpp
+++ b/src/hud_elements.cpp
@@ -417,7 +417,7 @@ void HudElements::frame_timing(){
         char hash[40];
         snprintf(hash, sizeof(hash), "##%s", overlay_param_names[OVERLAY_PARAM_ENABLED_frame_timing]);
         HUDElements.sw_stats->stat_selector = OVERLAY_PLOTS_frame_timing;
-        HUDElements.sw_stats->time_dividor = 1000.0f;
+        HUDElements.sw_stats->time_dividor = 1000000.0f; /* ns -> ms */
         ImGui::PushStyleColor(ImGuiCol_FrameBg, ImVec4(0.0f, 0.0f, 0.0f, 0.0f));
         double min_time = 0.0f;
         double max_time = 50.0f;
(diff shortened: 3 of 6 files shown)


flgx16:
The HUD information is updating way too fast on your branch

Joshua-Ashton:
Will take a look, thanks for testing

flgx16:
The problem is probably here: https://github.com/flightlessmango/MangoHud/blob/4aa147b91df60eeeb7abca5a443ca15be8fc218f/src/overlay.cpp#L86

params.fps_sampling_period is not in nanoseconds

Joshua-Ashton:
Fixed! :smile_cat:

flgx16:
Yep, that is fixed but it seems there's also a problem with the frametime graph

Joshua-Ashton:
@flightlessmango Not sure if anything needs to be done on your side cause I made some things use ns instead of being rounded down early.

If you want I can just round the frametime down to us before it gets logged to the file.

Joshua-Ashton:
Also, @flgx16 fixed that problem. Thanks for pointing it out :)

src/mesa/util/os_time.c
```diff
@@ -54,14 +54,21 @@
...
+#endif
 
 int64_t
 os_time_get_nano(void)
 {
 #if DETECT_OS_LINUX || DETECT_OS_BSD
 
    struct timespec tv;
-   clock_gettime(CLOCK_MONOTONIC, &tv);
+   clock_gettime(MANGOHUD_CLOCKTYPE, &tv);
```
baryluk, line 71: Technically speaking you should verify the return value, if it maybe returns `ENOTSUP` fallback to `CLOCK_MONOTONIC` or report an error or crash. I am not 100% sure it is worth it tho, as `CLOCK_MONOTONIC_RAW` is supported since kernel 2.6.28.

src/overlay_params.cpp
```diff
@@ -538,7 +538,7 @@ parse_overlay_config(struct overlay_params *params,
    params->enabled[OVERLAY_PARAM_ENABLED_cpu_load_change] = false;
    params->enabled[OVERLAY_PARAM_ENABLED_legacy_layout] = true;
    params->enabled[OVERLAY_PARAM_ENABLED_frametime] = true;
-   params->fps_sampling_period = 500000; /* 500ms */
+   params->fps_sampling_period = 500000000; /* 500ms */
```
baryluk, line 541: This is going to change the user visible API. Also, there is no point of having so detailed value provided by user, and it is easy to miss zeros with long values.

I think it would be good to heave the fps_sampling_period at least here in us just like before.

flgx16: It stays as it is to change from config afaik

baryluk: > It stays as it is to change from config afaik
> [e19c56f#diff-011194157ef3446f8d995bb62dca19a8e0f6ed503dad901bb87498d420797f36R135](https://github.com/flightlessmango/MangoHud/commit/e19c56ff2c488c8ac3f0c6afaeb9eae11fe13503#diff-011194157ef3446f8d995bb62dca19a8e0f6ed503dad901bb87498d420797f36R135)

You are right. My bad.

src/mesa/util/os_time.c
Joshua-Ashton: It's not worth it: nobody is using a kernel from 2008.

baryluk:
LGTM.

@Joshua-Ashton Did you see any particular issues with the current use of `CLOCK_MONOTONIC` that motivated you to change it to `CLOCK_MONOTONIC_RAW`? I agree it is better to use `CLOCK_MONOTONIC_RAW` and correct way do to it, but I doubt it would really show any issues in real world.

Joshua-Ashton:
Not specifically here but I have encountered 'interesting' clock pacing when using CLOCK_MONOTONIC before in other projects.

baryluk:
I would like to retract my LGTM.

After a more research and extra measurements, I think `CLOCK_MONOTONIC` is more accurate, especially for short durations.

The NTP adjustments mentioned in the manpages, and such, are actually required to maintain high accuracy of `CLOCK_MONOTONIC` not drifting, i.e. because of slightly clock frequency drifts.

There is no need for `CLOCK_MONOTONIC_RAW`.

Considering that there is no evidence of `CLOCK_MONOTONIC` by itself causing issues for mangohud, I would think this PR should be dropped.

Joshua-Ashton:
Your error thing doesn't really make sense. Monotonic Raw is not meant to provide accurate absolute time but rather accurate delta time. The rate shifting to compensate for NTP is precisely the issue this is meant to resolve.

baryluk:
> Your error thing doesn't really make sense. Monotonic Raw is not meant to provide accurate absolute time but rather accurate delta time. The rate shifting to compensate for NTP is precisely the issue this is meant to resolve.

Without NTP adjustement the monotonic raw doesn't provide accurate delta time.

Joshua-Ashton:
> > Your error thing doesn't really make sense. Monotonic Raw is not meant to provide accurate absolute time but rather accurate delta time. The rate shifting to compensate for NTP is precisely the issue this is meant to resolve.
> 
> Without NTP adjustement the monotonic raw doesn't provide accurate delta time.

IDK what to say to you because this is patently incorrect and you understand this situation backwards.

Joshua-Ashton:
> > Your error thing doesn't really make sense. Monotonic Raw is not meant to provide accurate absolute time but rather accurate delta time. The rate shifting to compensate for NTP is precisely the issue this is meant to resolve.
> 
> Without NTP adjustement the monotonic raw doesn't provide accurate delta time.

IDK what to say to this because it's patently untrue and you have this situation backwards.

Wine, for example, implements QueryPerformanceCounter with MONOTONIC_RAW which is used for delta timing in pretty much every Windows application.

baryluk:
> > > Your error thing doesn't really make sense. Monotonic Raw is not meant to provide accurate absolute time but rather accurate delta time. The rate shifting to compensate for NTP is precisely the issue this is meant to resolve.
> > 
> > 
> > Without NTP adjustement the monotonic raw doesn't provide accurate delta time.
> 
> IDK what to say to this because it's patently untrue and you have this situation backwards.
> 
> Wine, for example, implements QueryPerformanceCounter with MONOTONIC_RAW which is used for delta timing in pretty much every Windows application.
> 
> If your concern is rather fixing the clock being behind the actual rate of time, then it is pretty much guaranteed that whatever skew is happening from NTP will be much much worse than however many tenths of a ns that is behind/ahead.

Almost no CPU or motherboard has frequency oscillators accurate (absolute and temperature drift, and compensation) enough to be considered accurate. Most will have almost a second of drift per day (which is at least 10 microseconds per second). That is why the NTP adjustment is a thing.

The fact that Wine uses `MONOTONIC_RAW` for `QueryPerformanceCounter` is because that is mimicking the Windows semantic of that clock. And has nothing to do with what you should be using for proper measurements in time.

> The pace being consistent is what really matters anyway.

On a scale of less than a second, both counters will provide almost exactly same results, probably with less than 0.1% difference between them. Never more. However, `MONOTONIC` will provide more accurate measurements.

If you can provide actual data and comparison between the two, that demonstrate I am wrong, that would be great.

jackun:
Afaik RAW doesn't get slewed by NTP and constant step is more important that current time accuracy so +1 to RAW.

Also probably insignificent but....
> I have been using CLOCK_MONOTONIC_RAW for a long time but turned back to CLOCK_MONOTONIC since _RAW can be 5x to 10x slower than the other on some recent CPUs.
> Anyway, _RAW is better in quality so the choice depends on how frequently do you call gettime().
> Just for reference:
> Some latency numbers:
> on a Xeon E5 v4 (Broadwell): 750ns vs 120ns
> on a thinkpad x1 laptop (i5 Uxxx): 1200ns vs 170ns

hxxps://github.com/atdt/monotonic/issues/1#issuecomment-461610201

flgx16:
I also think RAW makes more sense

PR #443, opened 2020-12-21
 Prevent infinite recursion when using Mesa zink.

baryluk:
When using zink (`MESA_LOADER_DRIVER_OVERRIDE=zink`),
running any app (OpenGL or Vulkan), will trigger running
`glxinfo`. Unsetting of `LD_PRELOAD` is not sufficient,
because if using `zink`, it will load Vulkan, and load Vulkan
implicit layer for Mangohud, using `/usr/share/vulkan/implicit_layer.d/MangoHud.json`,
which will cause it to reenter this function in another process,
causing infinite recursion (until the Vulkan driver finally errors
out with too many VkDevices  / contexts created).
I have easily have seen recursion depths of 200+ until this happen,
and sometimes takes up to 2 minutes, with 2000 extra shell and `cut`
processes in the process tree.

Fix this by marking recursion entrance.

This makes glxmark start in less than 1 second, instead
about 60 seconds.

When at it us `glxinfo -B`, which is significantly faster
on zink, yet provides all information we need.

### DIFF
--- a/src/vulkan.cpp
+++ b/src/vulkan.cpp
@@ -611,8 +611,16 @@ void init_system_info(){
       trim(os);
       gpu = exec("lspci | grep VGA | head -n1 | awk -vRS=']' -vFS='[' '{print $2}' | sed '/^$/d' | tail -n1");
       trim(gpu);
-      driver = exec("glxinfo | grep 'OpenGL version' | sed 's/^.*: //' | cut -d' ' --output-delimiter=$'\n' -f1- | grep -v '(' | grep -v ')' | tr '\n' ' ' | cut -c 1-");
-      trim(driver);
+
+      const char* mangohud_recursion = getenv("MANGOHUD_RECURSION");
+      if (!mangohud_recursion) {
+         setenv("MANGOHUD_RECURSION", "1", 1);
+         driver = exec("glxinfo -B | grep 'OpenGL version' | sed 's/^.*: //' | cut -d' ' --output-delimiter=$'\n' -f1- | grep -v '(' | grep -v ')' | tr '\n' ' ' | cut -c 1-");
+         trim(driver);
+         unsetenv("MANGOHUD_RECURSION");
+      } else {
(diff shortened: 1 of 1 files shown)


baryluk:
I noticed that there is an environment variable for disabling implicit mangohud layer:

`$ cat ../../build/release/usr/share/vulkan/implicit_layer.d/MangoHud.json`

```json
{
    "file_format_version" : "1.0.0",
    "layer" : {
      "name": "VK_LAYER_MANGOHUD_overlay",
      "type": "GLOBAL",
      "api_version": "1.2.135",
      "library_path": "/usr/lib/mangohud/\$LIB/libMangoHud.so",
      "implementation_version": "1",
      "description": "Vulkan Hud Overlay",
      "functions": {
         "vkGetInstanceProcAddr": "overlay_GetInstanceProcAddr",
         "vkGetDeviceProcAddr": "overlay_GetDeviceProcAddr"
      },
      "enable_environment": {
        "MANGOHUD": "1"
      },
      "disable_environment": {
        "DISABLE_MANGOHUD": "1"
      }
    }
}
```


Maybe it could also work pretty well for this.

jackun:
> Maybe it could also work pretty well for this.

Probably. Also the setenv etc haven't been the best behaving either. Should probably get rid of them all and just add them with `env` to exec calls.

baryluk:
> > Maybe it could also work pretty well for this.
> 
> Probably. Also the setenv etc haven't been the best behaving either. Should probably get rid of them all and just add them with `env` to exec calls.

I agree with that using `execve` directly would be better, than messing with possibly other threads or sub-processes spawned concurrently.

baryluk:
BTW. This entire file is a mess, it incorrectly is using `getenv`, `unsetenv` and `setenv` all over the place, that violatates POSIX rules, and can crash. But that is already there. I will send separate patches to fix this.

But I would love this one to be merged first, as it is a small change.

PR #445, opened 2020-12-25
imgui wrap from WrapDB and build-source.sh rework

stephanlachnit:
Reworked for dear imgui wrap. The script is a bit longer than before, but it offloads the hard to Meson again.

Currently this doesn't work, I suspect due to some weird wrap-git error in Meson. Once https://github.com/mesonbuild/imgui/tree/1.79 is done for 1.80, it should work though.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,7 @@ build
 builddir
 __pycache__
 .vscode
-MangoHud*.tar.gz
+MangoHud*.tar.*
 pkg
 mangohud*.tar.*
 lib32-mangohud*.tar.*
--- a/build-source.sh
+++ b/build-source.sh
@@ -1,15 +1,36 @@
 #!/bin/sh
 
 VERSION=$(git describe --tags --dirty)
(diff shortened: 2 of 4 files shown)


stephanlachnit:
@jackun @flightlessmango imgui 1.80 got released, which means we can use the Wrap from WrapDB: https://wrapdb.mesonbuild.com/imgui

The official Wrap uses auto features, and sadly this can't be set per subproject yet (see https://github.com/mesonbuild/meson/issues/5320), so we have to disable each feature by hand for now.

I also reworked `build-source.sh` to work with the imgui Wrap. It now also creates two source tarballs, one with all files (including minhook, nvml.h and the vulkan headers), and the known DFSG tarball.

jackun:
This builds Dear ImGui as shared lib though

stephanlachnit:
Ah yeah sorry totally forgot to set that variable. Fixed now.

stephanlachnit:
Any chance this gets merged before 6.2? Else it would be really annoying to package.

PR #487, opened 2021-02-18
Don't show redundant frame timing on graph

iczero:
When frame times are enabled with the fps display (`frametime` option) and the frame timing graph is also enabled, the frame times are displayed twice. This removes the frametime display on the graph when the `frametime` option is active.

### DIFF
--- a/src/hud_elements.cpp
+++ b/src/hud_elements.cpp
@@ -551,15 +551,17 @@ void HudElements::wine(){
 void HudElements::frame_timing(){
     if (HUDElements.params->enabled[OVERLAY_PARAM_ENABLED_frame_timing]){
         ImGui::TableNextRow(); ImGui::TableNextColumn();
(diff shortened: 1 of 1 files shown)


iczero:
Tried basing off the `develop` branch but apparently the imgui meson wrap will not do static linkage. It's pretty much only one line anyways. Sorry for any inconvenience.

jackun:
> Tried basing off the `develop` branch but apparently the imgui meson wrap will not do static linkage.

It should, at least it does for me with meson 0.57. If deleting build dir and starting over doesn't fix it then maybe meson is too old?

iczero:
Oh, removing the build directory fixed it. Will rebase

gort818:
Just tested this, works as intended thanks, would you mind rebasing off the master branch

iczero:
Rebased against master and changed PR base.

flightlessmango:
I don't think this is necessary anymore since fb1cae9

PR #497, opened 2021-03-14
imgui 1.81

stephanlachnit:
I'm also looking into removing `src/gl/imgui_impl_opengl3.h` and `src/gl/imgui_impl_opengl3.cpp`, since they are provided in the Wrap. However, there are two problem at the moment: the Wrap doesn't allow for other GL loaders than GLEW, so that would need to be fixed, and in `src/gl/imgui_hud.cpp` `GetOpenGLVersion()` is called.


jackun:
`imgui_impl_opengl3.cpp` is a bit different with some "fixes"

stephanlachnit:
Anyway, this can still be merged, right?

PR #504, opened 2021-03-25
Update README.md

Brisse89:
Add Debian install instructions.


stephanlachnit:
One might want to add how to install the 32bit package: `sudo apt install mangohud:i386` (assuming 32bit is enabled).

PR #507, opened 2021-04-05
Fix Intel RAPL CPU power calculations

medusalix:
I ran into some problems while using the `cpu_power` option for the HUD. This PR fixes the following issues:

- [x] The `energy_uj` value read from RAPL is a 64-bit unsigned integer ([source](https://github.com/torvalds/linux/blob/e49d033bddf5b565044e2abe4241353959bc9120/drivers/powercap/intel_rapl_common.c#L156)), causing an overflow of the `energyCounterValue`.
- [x] The CPU power is evaluated on the first tick leading to a random power value because the `lastCounterValue` has not been set.
- [x] The power usage can become negative when the `energy_uj` value wraps around or resets.

### DIFF
--- a/src/cpu.cpp
+++ b/src/cpu.cpp
@@ -8,6 +8,7 @@
 #include <string.h>
 #include <algorithm>
 #include <regex>
+#include <inttypes.h>
 #include "string_utils.h"
 
 #ifndef PROCDIR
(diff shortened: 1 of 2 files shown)


flgx16 (approved):
LGTM. I didn't test the original code as much as the AMD code as I had to test on my laptop instead of my computer, so thanks for fixing it.

PR #531, opened 2021-05-25
Add support for multi gpu device stats

Zakhrov:
This allows setups with more than one GPU to see statistics from both GPUs. This is still pretty rough and has been tested only on Vulkan and not OpenGL or Direct3D. 
I am still pretty new to Vulkan programming and I thought this would be a good way to explore Vulkan device groups and Vulkan explicit Multi GPU (Whenever they become a thing)

### DIFF
--- a/src/hud_elements.cpp
+++ b/src/hud_elements.cpp
@@ -104,12 +104,14 @@ void HudElements::gpu_stats(){
     for(auto per_gpu_info: gpu_info){
          if (HUDElements.params->enabled[OVERLAY_PARAM_ENABLED_gpu_stats]){
         ImGui::TableNextRow();
-        const char* gpu_text;
+        
+        std::string gpu_text;
         if (HUDElements.params->gpu_text.empty())
-            gpu_text = "GPU";
+            gpu_text = "GPU " + per_gpu_info.second.deviceName;
         else
             gpu_text = HUDElements.params->gpu_text.c_str();
-        ImGui::TextColored(HUDElements.colors.gpu, "%s", gpu_text);
+        ImGui::TextColored(HUDElements.colors.gpu, "%s", gpu_text.c_str());
+        ImGui::TableNextRow();
         ImGui::TableNextCell();
         auto text_color = HUDElements.colors.text;
         if (HUDElements.params->enabled[OVERLAY_PARAM_ENABLED_gpu_load_change]){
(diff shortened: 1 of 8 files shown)


src/hud_elements.cpp
```diff
@@ -536,7 +545,7 @@ void HudElements::graphs(){
             arr.erase(arr.begin());
         }
 
-        HUDElements.max = gpu_info.memoryTotal;
+        HUDElements.max = gpu_info[0].memoryTotal;
```
flightlessmango, line 551: This seems like a typo from me? I assume that shouldn't be vram, it should be system ram

src/hud_elements.cpp
```diff
@@ -104,12 +104,14 @@ void HudElements::gpu_stats(){
     for(auto per_gpu_info: gpu_info){
          if (HUDElements.params->enabled[OVERLAY_PARAM_ENABLED_gpu_stats]){
         ImGui::TableNextRow();
-        const char* gpu_text;
+        
+        std::string gpu_text;
         if (HUDElements.params->gpu_text.empty())
-            gpu_text = "GPU";
+            gpu_text = "GPU " + per_gpu_info.second.deviceName;
```
flightlessmango, line 110: I don't think the entire device name is going to fit on the hud. I also prefer the device name that comes from pci.ids rather than vulkan

src/hud_elements.cpp
Zakhrov: I've pushed a fix for that. It is the total VRAM. The fix builds VRAM graphs for all GPUs

src/hud_elements.cpp
Zakhrov: The entire name does not fit. We could get the device code names (BONAIRE and CAPE VERDE in my case) instead of the full GPU name. The device name is actually coming from the pci.ids. I'm just storing the data from ```get_device_name``` which queries the PCI Ids
Screenshot_20210526_010124

flightlessmango: Thinking if it would be better to just name them "GPU0, GPU1" etc for now

flightlessmango:
This needs to be rebased on to `develop` branch

src/hud_elements.cpp
Zakhrov: I've tweaked the display in the new Pull request. I don't have that much experience with git rebases and messed up my commits. I ended up deleting my repository and adding my changes on top of the develop branch of a fresh fork.

PR #532, opened 2021-05-26
Expose stats of all GPUs

Zakhrov:
This is #531  rebased against develop.

### DIFF
--- a/src/hud_elements.cpp
+++ b/src/hud_elements.cpp
@@ -105,14 +105,17 @@ void HudElements::version(){
 }
 
 void HudElements::gpu_stats(){
-    if (HUDElements.params->enabled[OVERLAY_PARAM_ENABLED_gpu_stats]){
+    for(auto per_gpu_info: gpu_info){
+         if (HUDElements.params->enabled[OVERLAY_PARAM_ENABLED_gpu_stats]){
         ImGui::TableNextRow(); ImGui::TableNextColumn();
-        const char* gpu_text;
+
+        std::string gpu_text;
         if (HUDElements.params->gpu_text.empty())
-            gpu_text = "GPU";
+            gpu_text = "GPU " + per_gpu_info.second.deviceName;
         else
             gpu_text = HUDElements.params->gpu_text.c_str();
-        ImGui::TextColored(HUDElements.colors.gpu, "%s", gpu_text);
+        ImGui::TextColored(HUDElements.colors.gpu, "%s", gpu_text.c_str());
+        ImGui::TableNextRow();
         ImGui::TableNextColumn();
         auto text_color = HUDElements.colors.text;
         if (HUDElements.params->enabled[OVERLAY_PARAM_ENABLED_gpu_load_change]){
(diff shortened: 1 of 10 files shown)


jackun:
Next time just do `git push -f` to same branch, no need for new PR ;)

src/overlay.cpp
```diff
@@ -65,12 +65,18 @@ void update_hw_info(struct swapchain_stats& sw_stats, struct overlay_params& par
    // Save data for graphs
    if (graph_data.size() > 50)
       graph_data.erase(graph_data.begin());
-   graph_data.push_back(currentLogData);
+#ifdef _WIN32
```
jackun, line 68: This chunk is pointless it seems

src/vulkan.cpp
```diff
@@ -614,11 +638,11 @@ void init_system_info(){
       const char* mangohud_recursion = getenv("MANGOHUD_RECURSION");
       if (!mangohud_recursion) {
          setenv("MANGOHUD_RECURSION", "1", 1);
-         driver = exec("glxinfo -B | grep 'OpenGL version' | sed 's/^.*: //' | sed 's/([^()]*)//g' | tr -s ' '");
+         driver = exec("glxinfo -B | grep 'OpenGL version' | sed 's/^.*: //' | cut -d' ' --output-delimiter=$'\n' -f1- | grep -v '(' | grep -v ')' | tr '\n' ' ' | cut -c 1-");
```
jackun, line 641: Unrelated change from rebasing

src/vulkan.cpp
```diff
@@ -614,11 +638,11 @@ void init_system_info(){
...
          trim(driver);
          unsetenv("MANGOHUD_RECURSION");
       } else {
-         driver = "MangoHud glxinfo recursion detected";
+         driver = "MangoHud glxinfo recurssion detected";
```
jackun, line 645: Unrelated change from rebasing

src/hud_elements.cpp
```diff
@@ -105,14 +105,20 @@ void HudElements::version(){
 }
 
 void HudElements::gpu_stats(){
-    if (HUDElements.params->enabled[OVERLAY_PARAM_ENABLED_gpu_stats]){
+    for(auto per_gpu_info: gpu_info){
+         if (HUDElements.params->enabled[OVERLAY_PARAM_ENABLED_gpu_stats]){
         ImGui::TableNextRow(); ImGui::TableNextColumn();
-        const char* gpu_text;
```
jackun, line 110: Any reason for turning it into std::string instead?

Zakhrov: I was concatenating the name of the GPU. I've since put that on a separate hud element but we could keep it as a `std::string` in case we need to add GPU indices (GPU 1,GPU 2 etc)

src/vulkan.cpp
Zakhrov: Yeah I guess those are left over from merge conflict resolution

jackun:
Thought going with a bit more cpp-ish way, check https://github.com/flightlessmango/MangoHud/tree/libdrm-radeon
I've added a `show_all_gpus` (can has better name?) option to force drawing all gpus' stats.

SteavenGamerYT (approved):

PR #541, opened 2021-06-13
Add missing and rearrange options in .conf

Faalagorn:
Added missing options and rearranged the way they appear under legacy layout plus minor fixes:
* Added some default values after `=`
* Commented out options with default values
* Unified style

### DIFF
--- a/bin/MangoHud.conf
+++ b/bin/MangoHud.conf
@@ -23,7 +23,7 @@
 # custom_text_center=
 
 ### Display the current system time
-time
+# time
 
 ### Time formatting examples
 # time_format=%H:%M
--- a/bin/MangoHud.conf
+++ b/bin/MangoHud.conf
@@ -5,7 +5,7 @@
 
 ################ PERFORMANCE #################
 
-### Limit the application FPS. Comma-separated list of one or more FPS values (e.g. 0,30,60). 0 means unlimited (unless VSynced).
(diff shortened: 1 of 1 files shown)


Faalagorn:
This is mostly so that after enabling `legacy_layout=false` things should work as usual.

bin/MangoHud.conf
```diff
@@ -143,32 +166,28 @@ background_alpha=0.5
...
 
-### (Re)bind given framebuffer before MangoHud gets drawn. Helps with Crusader Kings III.
-# gl_bind_framebuffer = 0
+### (Re)bind given framebuffer before MangoHud gets drawn. Helps with Crusader Kings III
+# gl_bind_framebuffer=0
 
-### Don't swap origin if using GL_UPPER_LEFT. Helps with Ryujinx.
-# gl_dont_flip = 1
+### Don't swap origin if using GL_UPPER_LEFT. Helps with Ryujinx
+# gl_dont_flip=1
```
jackun, line 188: Spaces don't cost extra though

Faalagorn: I don't mind having extra space as long as the style is uniform, I think there was something that made me decide on having them removed, but I don't know what exactly. Do you want me to change them all to having spaces?

jackun: Eh, let it be

jackun:
In 87e25a3038f93be1ef7dc47a39ef056a769d7bd1

PR #605, opened 2021-09-24
Hud elements cat_file function

kokoko3k:
Acts like exec=cat /something.txt

...and I cloned as much as possible the existing exec implementation.


gort818:
Why not just use `exec=cat /something.txt` ?

kokoko3k:
> Why not just use `exec=cat /something.txt` ?

Efficiency.
Why spawning a new process every 1/2 second? :)

PR #618, opened 2021-10-28
build-source: exclude spdlog

stephanlachnit:
spdlog is available on most distros: https://repology.org/project/spdlog/versions

Thus, as with Vulkan headers, it would be appreciated if it could be removed from the DFSG tarball.


stephanlachnit:
@jackun @flightlessmango can you tag a release with this?

PR #631, opened 2021-11-10
Add GitHub compile workflow

stephanlachnit:
Just a simple workflow to check that MangoHud builds properly with new PRs. I rebased this against #151 and #627 as these change build options / dependencies.

You can view the result here: https://github.com/stephanlachnit/MangoHud/actions/runs/1443621148

I chose the `debian:testing` docker image as it is recent enough for all dependencies but not too bleeding edge to cause random failures. The `ubuntu-latest` (aka 20.04) image from GitHub e.g. doesn't have a recent enough version of Meson.
Once could of course argue that this is not really ideal as we probably also want to test building for Ubuntu 20.04 users, so this is open for discussion.

The workflow could also be extended with a 32bit build and other neat things like a static check from clang (the target is automatically created by meson if clang is present).


flightlessmango:
This is no longer needed

Issue #585, opened 2021-08-08
[Feature Request] Additional frametime metrics


kkartaltepe:
I was looking at `frame_timing` with mangohud and thought it would be nice to have a couple additional metrics to characterize the graph better.

1. Benchmark statistics available (either for the graphing period only or possibly for the full run)
2. An aggregated histogram (probably for the graphing period only). Im not sure what the intention of the `histogram` setting is, but it seems its just different way to render the existing frame_timings graph. I expected to see the frame timings aggregated into timing window buckets e.g. something like (pardon the ascii art)
```
     Frame time (ms) : count     distribution
       0 -> 1        : 3        |                                      |
       2 -> 3        : 0        |                                      |
       4 -> 7        : 211      |**********                            |
       8 -> 15       : 0        |                                      |
      16 -> 31       : 0        |                                      |
      32 -> 63       : 0        |                                      |
      64 -> 127      : 1        |                                      |
     128 -> 255      : 800      |**************************************|
```

The above example is just an exponentially increasing bucket size, but probably there are more appropriate ones given refresh rates.

One reasonable to me approach for benchmark stats would be to just use a 1002 element histogram to record all frame timings between 1ms and 1000ms, with over/under flow buckets for frames rendering faster than 1000fps or slower than 1fps allowing computation of benchmark percentiles at 1ms resolution and fast computation of a coarser histogram for display. Possibly allowing a hotkey to reset the data if support for the full run is considered useful.

If desired you could also add a few thousand more elements if people are interested in frame timings beyond 1000fps or below 1fps.

My main motivation is to see where frame hitches fall so knowing the 99%ile value would be helpful and its hard to see from the graph and doing it live is easier than going through the logging process. During some other runs where hitches seem multi-modal (some seem to stall for an entire frame while others stall for less time) it would be cool to just edit a histogram to show me (this is probably more cool than useful as compared to the live benchmark stats, at least to me).

floorcat:
I would like to have min max frame timings similar to mesa overlay:
![mesa-fs8](https://user-images.githubusercontent.com/91414655/136659179-7f243440-946c-4a36-86f8-01296ed410eb.png)
:

PR #659, opened 2021-12-20
[WIP] hud: Modify histogram to use buckets and report p50/p90/p99

kkartaltepe:
This makes the histogram option behave like a histogram instead of
simply rendering the same data as frame_timing without the histogram
option.

This is mostly for reference and personal use as is and does not match
some of the repo's coding standards.

This is a demonstration of a simple implementation of #585, if it seems reasonable and desirable upstream i would be happy to clean this up for better than my personal use. Alternately feel free to use any code in this PR for your own implementation.


kkartaltepe:
As its, its also far from beautiful but an example of how it renders is 
screen_20211219211941

PR #702, opened 2022-03-06
All the warnings

evelikov:
Tried building the project locally and was surprised that the in-tree PKGBUILD is busted.

Looking around let me to remove it, fixup the compiler warning, add -Werror to the existing CI and update mingw32.txt with a proper file.

As some point a pre-commit CI would be on the menu but for now this will do.


lukashoracek (changes requested):
I don't think the dependency and library folder Arch specific code in `build.sh` should be removed because it makes sure dependencies are installed when you build MangoHud through the script directly.

evelikov:
I don't think it makes sense to have Arch specifics in the build.sh script - as indicated by the PKGBUILD which used it ... it's basically unmaintained and/or broken.

In general, as Arch user you _really_ do not want random scripts to be installing packages on your system.

lukashoracek:
I agree that distro packages should be used to install MangoHud but installing through the script is useful for quicker MangoHud development and the dependency check is there already for other distros so IMO it makes sense to keep it there.

evelikov:
Strange, the script has cause damage than good in my case of very limited MangoHud development - bunch more patches incoming.

The fact that it reinvents the install/uninstall targets is also _very_ alarming and I'm inclined it ought to be reduced to a dummy meson/ninja wrapper.

Then again, I've reinstated said hunk since you feel strongly attached to it :-P

src/overlay_params.cpp
```diff
@@ -660,7 +660,6 @@ parse_overlay_config(struct overlay_params *params,
    if (!env || read_cfg) {
 
       // Get config options
-      bool has_config_file = parseConfigFile(*params);
```
jackun, line 663: This is half-send revert of https://github.com/flightlessmango/MangoHud/commit/beecde110b54badbb7ba4a8255c265a0e506d737
`parseConfigFile(*params);` should stay probably.

.github/workflows/build-package.yml
```diff
@@ -41,7 +41,7 @@ jobs:
       - name: Build release package
         run: |
           ./build-source.sh
-          ./build.sh build package release
+          ./build.sh CONFIGURE_OPTS="-Dwerror=true" build package release
```
jackun, line 44: Just `...build -Dwerror=true ... ` then.

src/overlay_params.cpp
evelikov: Would be better to finalise the revert, wouldn't it? One can always cherry-pick the original patch as needed in the future.

.github/workflows/build-package.yml
evelikov: Keep in mind that the `-Dwerror` option is meson, so it need to be propagated within the `build.sh` file. Are you OK with that or have something else in mind?

jackun: `build` "command" passes dash-prefixed stuff to meson itself. Well, it _should_ anyway.

evelikov: Indeed it does pass them along, thanks.

Issue #53, opened 2020-03-06
Vega 3 gets 0% use


BNandroid:
I use an AMD Athlon 200GE APU with Vega 3 Graphics and my gpu load is 0% :/

jackun:
What does `cat /sys/class/drm/card*/device/gpu_busy_percent` output?

BNandroid:
> What does `cat /sys/class/drm/card*/device/gpu_busy_percent` output?

Invalid Argument

Do I need to install any packages?

jackun:
For some reason `gpu_busy_percent` doesn't support Vega 3 then. 
Can you try [radeon-profile](https://github.com/marazmista/radeon-profile) or radeontop? If these work then using ioctl instead might work.

theron29:
I have exactly the same issue on my notebook with Vega10.
> cat /sys/class/drm/card*/device/gpu_busy_percent 
also returns Invalid Argument.

Using Oibaf drivers on Ubuntu 19.10 with ACO enabled.

BNandroid:
> For some reason `gpu_busy_percent` doesn't support Vega 3 then.
> Can you try [radeon-profile](https://github.com/marazmista/radeon-profile) or radeontop? If these work then using ioctl instead might work.
> 
> Maybe adding amdgpu.ppfeaturemask to kernel command line might help like `amdgpu.ppfeaturemask=0xfffd7fff`

I will test this later, thanks 😁

flightlessmango:
@BNandroid did you try it and if so did it work?

ViNi-Arco:
hi, it happens to me too, I have Nvidia card, see does not show the use of the GPU nor the processor temperature:

![2020-03-09-163935_1440x900_scrot](https://user-images.githubusercontent.com/57958460/76463703-a28a4f80-63c2-11ea-9ea1-f5da9368047d.png)

BNandroid:
> @BNandroid did you try it and if so did it work?

I tried everything, still 0%: /

flightlessmango:
@ViNi-Arco your 0% gpu load is probably is probably because your Nvidia card is not supported in NVML, what GPU is it? why the cpu temp isn't showing I don't know, what CPU is it?

ViNi-Arco:
Good morning @flightlessmango, I have an EVGA GTX 650, it has NVML support, because in MSI Afterburner I can see the GPU usage and the CPU Temp, my Processor is the E8400 rev.E0

[...]

PR #703, opened 2022-03-06
Remove libdrm_amdgpu and other unnecessary code 

evelikov:
This series removes the a bunch of dead or no longer needed code. In particular:

 - libdrm_amdgpu has two better alternatives in-tree - individual and the gpu_metrics sysfs nodes
   - bonus now we properly track/open only the required ones
- ifdef spaghetty in keybinds code
- the find_folder() API - we can use ls() instead
- excessive pipelines when fetching basic data

The vendor/device handling is still a bit iffy - I might be sorting that out next.

Note: this has been only lightly tested, so please give it a spin on your end and report issues.


evelikov:
Been using this for a few days without issues. Just some 600+ loc removed

@flightlessmango @jackun any takers?

Faalagorn:
I gave it a quick try on my RX 580, so far no issues at the first glance, I'll keep on running with that to see if I encounter any issues down the road.

flightlessmango:
It looks good to me, I would just like @jackun to sign off as well

src/overlay.cpp
```diff
@@ -639,57 +638,35 @@ void init_gpu_stats(uint32_t& vendorID, uint32_t reported_deviceID, overlay_para
...
+               amdgpu.memory_clock = fopen((path + dir + "/freq2_input").c_str(), "r");
             if (!amdgpu.temp)
-               amdgpu.temp = fopen((path + tempFolder + "/temp1_input").c_str(), "r");
+               amdgpu.temp = fopen((path + dir + "/temp1_input").c_str(), "r");
             if (!amdgpu.power_usage)
-               amdgpu.power_usage = fopen((path + tempFolder + "/power1_average").c_str(), "r");
+               amdgpu.power_usage = fopen((path + dir + "/power1_average").c_str(), "r");
 
-            vendorID = 0x1002;
             break;
```
jackun, line 664: Tangling `break`?

jackun:
Do Vega APUs support `gpu_metrics`, and why libdrm etc? At least Vega64 dgpu doesn't seem to.

src/overlay.cpp
```diff
@@ -639,57 +638,35 @@ void init_gpu_stats(uint32_t& vendorID, uint32_t reported_deviceID, overlay_para
...
             SPDLOG_DEBUG("Using gpu_metrics");
          }
-#ifdef HAVE_LIBDRM_AMDGPU
-         else {
-            int idx = -1;
-            //TODO make neater
-            int res = sscanf(path.c_str(), "/sys/class/drm/card%d", &idx);
-            std::string dri_path = "/dev/dri/card" + std::to_string(idx);
-
-            if (!params.enabled[OVERLAY_PARAM_ENABLED_force_amdgpu_hwmon] && res == 1 && amdgpu_open(dri_path.c_str())) {
```
jackun, line 649: Then no need for `force_amdgpu_hwmon` param

src/overlay.cpp
evelikov: Yup should be after the `}` - fix coming in a moment.

evelikov:
> Do Vega APUs support `gpu_metrics` (a.k.a why libdrm etc used in first place)? At least Vega64 dgpu doesn't seem to. e: vega12, vega20, renoir, vangogh, navi10, navi21(?), MI cards, maybe more... so only APUs since 2020?

AFAICT the standalone nodes - gpu_busy_percent, temp1_input and others, are present on all GPUs. So even if the composite `gpu_metrics` node is missing, all GPU will still have basic data.

That said, looks like I've forgot a check - currently the `gpu_metrics` node will be omitted if `gpu_busy_percent` is missing. It seems very unlikely to happen, but still not the right thing to do. Will add extra patch on top.

src/overlay.cpp
evelikov: Indeed - will drop it.

jackun:
IIRC `gpu_busy_percent` gave "Invalid Argument" on Vega 3 etc. unless patched since https://github.com/flightlessmango/MangoHud/issues/53

evelikov:
Sounds like something which should be reported and fixed in the amdgpu driver. Adding/maintaining a complete code-path for such a bug that does not sound palatable IMHO.

jackun (approved):

evelikov:
Doing a quick test shows that something has gone off.
Quickly flipping to draft - so this doesn't go in and break the world

evelikov:
The `gpu_busy_percent check` that the last commit removed had a _very_ misleading meaning - it was used to distinguish between `card0` and `card0-DP-1` nodes. See the commit description (and inline comment) for details.

jackun:
> to distinguish between the card node (cardX) and the
card output node (cardX-output-foo).

More of a is using `amdgpu` check but filters those (cardX-output-foo) out too :P

evelikov:
Fwiw there's another dozen or so patches series sitting on top of this work - https://github.com/flightlessmango/MangoHud/pull/707

jackun:
APUs need newer (v)bioses, I guess: https://github.com/torvalds/linux/blob/d888c83fcec75194a8a48ccd283953bdba7b2550/drivers/gpu/drm/amd/pm/powerplay/hwmgr/smu10_hwmgr.c#L1291

PR #707, opened 2022-03-13
Various mangoapp fixes, et al

evelikov:
While playing around with the mangoapp I've noticed a couple of serious quirks:
 - we destroy the glfw window, while checking the destroyed object - it's a miracle we don't crash
 - to prevent visual glitches on window resizes, we render 3x times

The former was trivial to resolve - just iconify/restore the window.

Fixing the resize/render was not so straight-forward, since we seemingly leak/duplicate a lot of state across components and modify it in fairly convoluted manner.
So this is not resolved, yet things should be tad easier to reason - with API cleaned up and immutable state annotated as such.

The series build on top of my earlier MR - https://github.com/flightlessmango/MangoHud/pull/703

### DIFF
--- a/src/app/control.cpp
+++ b/src/app/control.cpp
@@ -3,7 +3,6 @@
 #include "mangoapp.h"
 
 void help_and_quit() {
-    // TODO! document attributes and values
     fprintf(stderr, "Usage: mangohudctl [set|toggle] attribute [value]\n");
     fprintf(stderr, "Attributes:\n");
     fprintf(stderr, "   no_display      hides or shows hud\n");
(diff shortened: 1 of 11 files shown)


evelikov:
@flightlessmango @jackun any comments?

jackun:
`#ifndef MANGOAPP_LAYER` in overlay.h lost its `#endif` somehow, looks like

src/cpu.cpp
```diff
@@ -403,6 +403,7 @@ static bool find_temp_input(const std::string path, std::string& input, const st
     return false;
 }
 
+#ifndef MANGOAPP
```
jackun, line 406: @flightlessmango keeps flip-flopping :D unnecessary now (but the value of find_fallback_temp_input itself is suspect).

PR #710, opened 2022-03-18
shell completion for mangohud & mangohudctl

riscygeek:
This is a small PR for shell completions (currently only bash).
It adds a new meson build option ```-Dbashcompletiondir=```.
Unfortunately, I've got neither experience with zsh nor with fish completions, but it should be pretty simple to add.
Just look into `data/meson.build` and `data/shell-completion/bash/meson.build` for reference.


evelikov:
Assuming this gets anywhere, I can zsh completions o/

riscygeek:
> Assuming this gets anywhere, I can zsh completions o/

Yeah, that would be good.

meson_options.txt
```diff
@@ -2,6 +2,7 @@ option('glibcxx_asserts', type : 'boolean', value : false)
 option('use_system_vulkan', type : 'feature', value : 'disabled', description: 'Use system vulkan headers instead of the provided ones')
 option('use_system_spdlog', type : 'feature', value : 'disabled', description: 'Use system spdlog library')
 option('vulkan_datadir', type : 'string', value : '', description: 'Path to the system vulkan headers data directory if different from MangoHud\'s datadir')
+option('bashcompletiondir', type: 'string', description: 'Directory for bash completion script ["no" disables]')
```
evelikov, line 5: IMHO the extra toggle is an overkill. The completion files are not even 50 lines in total, so I'd just install them always.

riscygeek: This option specifies _where_ to put the bash completions.
It may not be always in `${prefix}/share/bash-completion`.
It just happens to allow disabling it.
I copied it from systemd's meson config.

evelikov: The existing logic will work with virtually any linux distro, even ones like Yocto. So in practical terms, the toggle is used for disabling the lot.

The fragment looked oddly familiar - thanks for confirming it's systemd origin.

data/shell-completions/bash/meson.build
```diff
@@ -0,0 +1,15 @@
...
+  if bash_completion.found()
+    bashcompletiondir = bash_completion.get_variable(pkgconfig: 'completionsdir', default_value: bashcompletion_defaultdir)
+  else
+    bashcompletiondir = bash_completion_defaultdir
+  endif
+endif
+
+if bashcompletiondir != 'no'
+  install_data('mangohud',    install_dir: bashcompletiondir)
+  install_data('mangohudctl', install_dir: bashcompletiondir)
```
jackun, line 14: Can you add `install_tag: 'scripts'` for these `install_data`s?
Or some better tag though I already tagged `mangohud` with it :P

PR #719, opened 2022-03-30
Check if amdgpu gpu_metrics can be read and do some sanity-checks

jackun:


evelikov:
Fwiw adding support for more metrics is rather trivial. I have some patches for that already.
Will try to find time and rebase/open a MR in the upcoming days.

PR #724, opened 2022-04-06
add functions to check if a process is running

azadad96:
Now `bool process_running(const char *process_name)` and `int process_id(const char *process_name)` can be used to check if a process containing a certain string in the process' command is running.
This could for example be used to determine if OBS is running and show the user an indicator.

Please not that I am not accustomed to using meson, therefore I might have added the file's name to the wrong place. It still builds and works however.

### DIFF
--- a/src/meson.build
+++ b/src/meson.build
@@ -83,6 +83,7 @@ if is_unixy
     'battery.cpp',
     'control.cpp',
     'gamepad.cpp',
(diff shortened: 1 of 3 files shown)


flightlessmango:
I don't see much use for this as a feature. Someone might be interested if OBS is recording or not, but I doubt it's useful to know if the app is on or not

PR #735, opened 2022-04-19
Introduce FCAT markers

PMunkes:
This is a first implementation of the Nvidia FCAT markers into MangoHud. I originally wrote this for the Digital Foundry Steam Deck review, but I never got around to actually upstream the code. 
The code is in pretty rough shape at the moment, since it has not been hooked up to any settings.

### DIFF
--- a/src/overlay.cpp
+++ b/src/overlay.cpp
@@ -14,6 +14,7 @@
 #include "gpu.h"
 #include "memory.h"
 #include "timing.hpp"
+#include "fcat.h"
 #include "mesa/util/macros.h"
 #include "battery.h"
 #include "gamepad.h"
@@ -38,6 +39,7 @@ float g_overflow = 50.f /* 3333ms * 0.5 / 16.6667 / 2 (to edge and back) */;
 string gpuString,wineVersion,wineProcess;
 uint32_t deviceID;
 bool gui_open = false;
+bool fcat_open = false;
 struct benchmark_stats benchmark;
 struct fps_limit fps_limit_stats {};
 ImVec2 real_font_size;
@@ -49,6 +51,7 @@ bool gpu_metrics_exists = false;
 bool steam_focused = false;
 vector<float> frametime_data(200,0.f);
 int fan_speed;
+fcatoverlay fcatstatus;
 
 void init_spdlog()
 {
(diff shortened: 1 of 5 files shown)


PMunkes:
I am currently unsure of 2 things:

1. Should the positions of other overlays be adjusted if the FCAT overlay is active?

I'm not sure if this is necessary since this would be a benchmark tool for professional users.

2. How should this be exposed to the users?

I think the optimal way would be a set of settings:

1. A customizable hotkey to enable the FCAT overlay.
2. The screen edge to overlay this over.
3. The default state of the overlay. This would enable a fire-and-forget approach, where the overlay is active from the start of the application.
4. A cusomizable width? (This could be useful for other applications, such as speedrunning, where a very small 2px overlay could be useful to prove consistency.)

xcasxcursex:
I'm using FCAT markers for game analysis on windows, and planning a migration to linux, so I'm really pleased to see this PR, thanks mate. 

I'm a rookie at this, so my thoughts should be taken with a grain of salt, but I'd prefer for nothing else to change aside from the marker being added (so, leave other overlays in the same position). Hope that helps.

PMunkes:
This code is now ready to be reviewed. It's hooked up to the regular configuration option infrastructure and can dynamically adjust it's settings.
The FCAT markers will draw on top of everything else and (currently) ignore the settings to disable the HUD, this is intentional behaviour.

PMunkes:
Here is a build with the overlay included. The usual warnings about running code from strangers should still apply.
[MangoHud-0.6.6.1.r218.g57b3d9f.tar.gz](https://github.com/flightlessmango/MangoHud/files/8747463/MangoHud-0.6.6.1.r218.g57b3d9f.tar.gz)

PMunkes:
@flightlessmango @jackun any comments or change requests?

src/overlay_params.h
```diff
@@ -225,6 +228,7 @@ struct overlay_params {
    unsigned media_player_color;
    unsigned table_columns;
    bool no_small_font;
+   bool fcat;
```
jackun, line 231: Unused probably, already in `overlay_params::enabled[OVERLAY_PARAM_ENABLED_fcat]`

jackun:
Probably squash it to single commit.
Global `fcatoverlay fcatstatus` probably gives incorrect colors with multi-swapchain apps, but don't know if there's any you'd use fcat with anyway.

src/overlay_params.h
PMunkes: Yeah, this is due to me misunderstanding how the parameters work. I'll change it.

PMunkes:
I don't know if multi-swapchain apps would be an issue for MangoHud, but this can be something that can be resolved if it ever comes up. For triple-buffered VSYNC, skipped colours are actually the intended behaviour FWIW.

PMunkes:
@jackun every rendered frame should have the next colour in the sequence. If this results in skipping some that were not displayed, that would be intentional.

jackun:
> multi-swapchain

Ah, `get_next_color()` could just take the `swapchain_stats& data` from `render_imgui()`, probably enough.

src/overlay.cpp
```diff
@@ -568,6 +571,22 @@ void render_imgui(swapchain_stats& data, struct overlay_params& params, ImVec2&
...
+       fcatstatus.update(&params);
+       auto window_corners = fcatstatus.get_overlay_corners();
+       auto p_min=window_corners[0];
+       auto p_max=window_corners[1];
+       auto window_size= window_corners[2];
+       ImGui::SetNextWindowPos(p_min, ImGuiCond_Always);
+       ImGui::PushStyleVar(ImGuiStyleVar_WindowPadding, ImVec2(0,0));
+       ImGui::SetNextWindowSize(window_size);
+       ImGui::Begin("FCAT", &fcat_open, ImGuiWindowFlags_NoDecoration| ImGuiWindowFlags_NoBackground);
+       ImGui::GetWindowDrawList()->AddRectFilled(p_min,p_max,fcatstatus.get_next_color(&data),0.0);
```
jackun, line 586: Nitpick, why not just by ref

PR #771, opened 2022-05-23
Update Vulkan layer api version to 1.3.211

EdoardoLuciani:
Hi everyone! While trying to run a Vulkan 1.3 application with Mangohud, it does not initialise and no debug output is written to the screen. I fixed the issue by bumping up the Vulkan layer API version in my local configuration.

In the commit, I applied the same modifications I did to my local files.


EdoardoLuciani:
Closing since a commit has been done that precisely addressed this issue

Issue #775, opened 2022-05-31
Gamescope External Overlay atom no longer released


trigg:
As per this commit : https://github.com/flightlessmango/MangoHud/commit/74e83c0bbafd048dce820a117fd6c7c6cb425c43

Previously on Steam Deck when setting 'Performance Overlay Level' to 0 the overlay window would be destroyed and other applications could draw to the external overlay layer.

As of this commit it is now impossible for any other app to draw into the overlay layer as mangoapp is run on startup and never releases it.

Related issue on my end:
https://github.com/trigg/Discover/issues/253

trigg:
I've made a [basic PR](https://github.com/Plagman/gamescope/pull/510) to allow windows to unset themselves from the external overlay slot. If this gets accepted all you'd need to do is toggle `GAMESCOPE_EXTERNAL_OVERLAY` to 1 and 0 for show and hide

trigg:
https://www.reddit.com/r/SteamDeck/comments/v50c29/part_of_my_screen_just_died_to_my_surprise_it/

https://www.reddit.com/r/SteamDeck/comments/v2v6c3/steam_deck_occasionally_doing_this_only_way_to/

My suspicion is that we've got a few bugs with an undrawn window being used in the compositing.

trigg:
![20220616160253_1](https://user-images.githubusercontent.com/535772/174105333-cfbe385d-4147-4187-b4e8-e6f594f2eed8.jpg)
![20220616162813_1](https://user-images.githubusercontent.com/535772/174105738-090b8607-fb67-4eaf-8f1d-ccd6c8ee9097.jpg)

Switching on & off sometimes leaves behind old content.
Tested in Beta & 'Main' os channels.

trigg:
https://www.reddit.com/r/SteamDeck/comments/vt5889/black_rectangle_on_deck/

PR #781, opened 2022-06-05
Add and remove XAtom in gamescope mode

trigg:
Intended fix for #775 

Removes and adds XAtom when `no_display` is changed. This will stop artifacting from an undrawn overlay and also release the layer for other overlays to draw to.


flightlessmango:
I've been unable to test this change. Adding discover as a non steam game works fine, however discord never starts, stalls and can't exit

flightlessmango:
It seems to work well with just mangoapp. That's probably good enough.

PR #808, opened 2022-07-26
changing Logger class from struct pointer to 3 variables

tisim:
On my PC after compiling mangohud, logging with keys showed random behaviour.
Sometimes NaN values, sometimes no output file. sometimes same values for all percentiles and average, but then only one line in output file.

Tracking down the issue, it seemed that on my machine, at least running Gothic1 with mods (32 bit), that the pointer to the overlay param struct given to the logger class does not seem to keep its values during the whole runtime.
Probably due to the struct behind it losing its scope.

Thus instead of passing the pointer to the whole struct I changed passing the 3 necessary values only and storing them directly instead of just a pointer to the struct.

### DIFF
--- a/src/overlay_params.cpp
+++ b/src/overlay_params.cpp
@@ -835,7 +835,6 @@ parse_overlay_config(struct overlay_params *params,
    else
       HUDElements.text_column = 1;
 
-   if(logger && logger->m_params == nullptr) logger.reset();
    if(!logger) logger = std::make_unique<Logger>(HUDElements.params);
    if(params->autostart_log && !logger->is_active())
       std::thread(autostart_log, params->autostart_log).detach();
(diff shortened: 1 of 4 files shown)


AdelKS:
I had the same issue on my system. And this PR fixed it. I wonder why doesn't it affect everyone. Thanks!

src/overlay_params.cpp
```diff
@@ -835,7 +835,6 @@ parse_overlay_config(struct overlay_params *params,
    else
       HUDElements.text_column = 1;
 
-   if(logger && logger->m_params == nullptr) logger.reset();
```
flightlessmango, line 838: Why did you decide to remove this line?
Without this the logger will never reinitialize and we only save the param values when the logger initializes.
So if you were to change `output_path` while running mangohud, it wouldn't actually change in the logger

flightlessmango:
This has been merged here 732439b8e68ed68166d9684a776cd9d5521b4168

PR #811, opened 2022-08-01
Updated PKGBUILD to honor the user set compiler using CC and CXX environment variables.

Billli11:
Update the PKGBUILD file to check CC and CXX variables.

Then add "-m32" to variables if those are set, and fallback to GCC as the compiler if not.

### DIFF
--- a/pkgbuild/PKGBUILD
+++ b/pkgbuild/PKGBUILD
@@ -53,8 +53,8 @@ build() {
     ${_build_args} -Dmangoapp=true -Dmangohudctl=true
 
   ninja -C build64
-  export CC="gcc -m32"
-  export CXX="g++ -m32"
+  export CC="${CC:-gcc} -m32"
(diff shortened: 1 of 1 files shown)


pkgbuild/PKGBUILD
```diff
@@ -53,8 +53,8 @@ build() {
     ${_build_args} -Dmangoapp=true -Dmangohudctl=true
 
   ninja -C build64
-  export CC="gcc -m32"
-  export CXX="g++ -m32"
+  export CC="${CC:-gcc} -m32"
+  export CXX="${CXX:-g++} -m32"
```
flightlessmango, line 57: Don't we want this to affect the 64bit build as well?

Billli11:
It's fine becase the problem is CC and CXX is explicitly set when building the 32bit version.
Those two variable is not set in PKGBUILD before so that when building 64bit is already using user set compiler.

PR #830, opened 2022-08-28
Use meson to handle multilib install and uninstall

AdelKS:
Hello,

Since I saw that meson automatically handles library paths according the host machine, it should handle properly MangoHud's multilib building.

In this PR:
- Rely on meson to determine whether `/usr/lib`, `/usr/lib64`... etc should be used for both x86 and x86_64 builds.
- Use `meson install` to install directly into the `/usr` prefix and without specifying in `build.sh` which files to install.
- Use meson's list of installed files for uninstalling.

I feel like there are some final touches to do, but I cannot see them for now. Otherwise this needs to be tested on Arch, RPM and DEB distros to confirm everything is alright. It works on Gentoo.

Feedback welcome!

Adel.


build.sh
```diff
@@ -144,133 +146,106 @@ configure() {
...
-            ;;
-        esac
-    done
-
-    if [ "$MACHINE" != "x86_64" ]; then
-        # Native libs
-        DEFAULTLIB=lib64
+    # The file build/meson-$MACHINE/install-log.txt contains all the files installed by meson
+    if [[ -f $INSTALL_LIST_NATIVE ]]; then
+        $SU_CMD xargs --arg-file=$INSTALL_LIST_NATIVE --verbose rm -rf
```
AdelKS, line 209: Now that I think of it, this may be dangerous, I think it's better to make the files `$INSTALL_LIST_NATIVE` and `$INSTALL_LIST_MULTILIB` root owned.

eli-schwartz: Wouldn't an attacker just edit build.sh instead then? Or intercept install-log.txt / write additional lines to it?

AdelKS: Fair point. If one's machine is running malware as the same user, it's kinda done for. 

Although my suggestion may still be useful for
- not losing the file to `git clean -ffdx` and co ? 
- it kinda belongs to root, conceptually, as the files that got installed are root owned.

🤔

AdelKS: Ah yeah, rebuilding over the already made meson build folder causes issues

```
PermissionError: [Errno 13] Permission denied: 'meson-private/../meson-logs/install-log.txt'
```

jackun:
`build.sh` is used for building release package primarily so needs to line-up with `mangohud-setup.sh` (a.k.a no meson to configure stuff there).
But these should probably be split as workflow and DIY :P

AdelKS:
You got me completely lost, where is this `mangohud-setup.sh` script ? What does it do ? How come something doesn't use the build system of  MangoHud  ?

jackun:
[Lolwut](https://github.com/flightlessmango/MangoHud/blob/master/bin/mangohud-setup.sh) and every package in [release](https://github.com/flightlessmango/MangoHud/releases)

AdelKS:
Oh I get it better now, the script that installs from the binary tarball. The meson approach of this PR could be transferred there too: 
1. Use `DESTDIR="$PWD/build/release" ninja -C build/meson-$MACHINE install` and `DESTDIR="$PWD/build/release" ninja -C build/meson-x86 install`
2. tar `$PWD/build/release` as-is and just untar that into the host system root folder.
3. this approach needs to keep a list of the installed files to be uninstalled, though.

But basically this is re-implementing a package manager. I suppose we do this to avoid making an Arch, DEB and RPM package.

AdelKS:
Nevermind, it needs the lib32 detection logic still...

AdelKS:
> build.sh is used for building release package primarily so needs to line-up with mangohud-setup.sh (a.k.a no meson to configure stuff there).

Yeah now that I had a quick look, I broke that workflow I think, I will check more thoroughly and see if I can update that with the changes done here.

> But workflow and DIY should use separate build scripts anyway probably :P

I am for using meson when installing on own machine. I started this PR without the tar+install script approach in mind (I wasn't aware that it even existed)

AdelKS:
Okay, so there are quite a few install scripts and I do have the energy to go through all of them. So how about I do this: 
- I make the script called `install.sh` and is to only be used for local installs with meson/compilation.
- Leave the rest untouched xD
Is that a good start, but that would be the fifth build/install script xD

PR #831, opened 2022-08-28
Add mangoplot script

AdelKS:
Hello guys, as I started benchmarking with `log_interval=0` to have accurate percentiles (MangoHud otherwise simply saves to disk 1 frame every log_interval), the produced files are too big to be handled by the website. A local way to plot and compare between benchmarks was lacking, so here it is.

This PR introduces a python script that enables comparing visually between benchmarks, and more accurately than what the website displays. It depends on `matplotlib` and needs a graphical back-end, I randomly chose `PyQt5`, we can decide on something else.

Here's an example output of the script (on a set of benchmarks I did)

benchmarks

It shows the distribution of the frame rates (`1 / (frame time)`), the 50%, the 1% and 0.1% percentiles. The `n`% percentiles are not exactly the `n`% lows: the `n`% lows involves averaging all the frame rates below the `n`% percentile.

Visualizing the distribution of the frame times enables seeing: 
- More nuances between benchmarks to properly assess what one prefers in one's own use-case
- That a benchmark is not reproducible yet, if e.g. a rebench doesn't visually have the same distribution

we can even switch to this view on the website, maybe, visualizing the frames is pretty, but pretty un-useful, except maybe to confirm that two benchmarks are related by seeing that they have the same trends.

The script probably needs some polishing and maybe some extra features like:
- Give which percentiles to display
- Accept a list of benchmark files instead of only a folder.
- Set an `fps_min` and an `fps_max` to crop the view (although it can be done with user input on the `matplotlib` window)

Feedback welcome !


build.sh
```diff
@@ -197,6 +197,7 @@ uninstall() {
     rm -fv "/usr/share/vulkan/implicit_layer.d/MangoHud.x86.json"
     rm -fv "/usr/share/vulkan/implicit_layer.d/MangoHud.x86_64.json"
     rm -fv "/usr/bin/mangohud"
+    rm -fv "/usr/bin/mangoplot"
```
jackun, line 200: Added these to mangohud-setup.sh too, probably.

AdelKS: done!

meson.build
```diff
@@ -8,7 +8,13 @@ project('MangoHud',
...
 cc = meson.get_compiler('c')
 cpp = meson.get_compiler('cpp')
-prog_python = import('python').find_installation('python3', modules: ['mako'])
+
+python_deps = ['mako']
+if get_option('mangoplot')
+  python_deps += ['matplotlib', 'PyQt5']
+endif
+
+prog_python = import('python').find_installation('python3', modules: python_deps)
```
jackun, line 17: Is this check even necessary for building? Shebang in py runs first found python3 anyway.

AdelKS: It's not, just for packagers to see the dependency x)

stephanlachnit: Should be optional then

AdelKS: I am not sure I understand with "should be optional", installing `mangoplot` is already optional (but default on) and the dependency is removed if installing `mangoplot` is disabled.

stephanlachnit: I.e. `check_runtime_dependencies`. When building packages in distributions usually runtime deps are not installed to reduce package build times (relevant if you do massive rebuilds, period CI builds, etc)

AdelKS: where is this `check_runtime_dependencies` ? x)

stephanlachnit: No, you need to make it a build option yourself

NoXPhasma:
I just tried the script and it does a good job. But two suggestions:

1. The script currently truncates the file name, which is used as title for the benchmark bar, at the first `.`. We already know that it is a `csv` file, we should just truncate the file extension.
2. The script currently breaks when the summary files are present in the folder as well. It would be better if we skip those, or any `csv` file which fails to be parsed.

PR #849, opened 2022-10-26
Lower the minimum required version of meson to v0.54.0

guihkx:
Unless strictly necessary, the minimum required version of meson should always be set to the oldest version possible. This is done to ensure compatibility with systems where upgrading meson is not possible.

For example, on specific branches of Flatpak SDKs: https://github.com/flathub/org.freedesktop.Platform.VulkanLayer.MangoHud/pull/19#issuecomment-1291596497

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,6 +1,6 @@
 project('MangoHud',
   ['c', 'cpp'],
-  version : 'v0.6.8',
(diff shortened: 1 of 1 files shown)


guihkx:
Sorry, I made a mess and meson 0.60 is actually required because of `install_tags`... 😅

PR #873, opened 2022-12-28
Fixup device detection on intel GPUs

kkartaltepe:
This resolves missing GPU names due to deviceID only being computed for AMD gpus. ~~It also fixes the broken vendor check before the AMD specific path checks.~~

### DIFF
--- a/src/app/main.cpp
+++ b/src/app/main.cpp
@@ -276,6 +276,8 @@ int main(int, char**)
     if (deviceName.find("Radeon") != std::string::npos
     || deviceName.find("AMD") != std::string::npos){
         vendorID = 0x1002;
+    } else if (deviceName.find("Intel") != std::string::npos) {
+        vendorID = 0x8086;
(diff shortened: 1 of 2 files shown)


kkartaltepe:
I dont have an AMD card to test with, but I presume that they also expose the vendor in the pci device path so this new vendor check should work on both.

flightlessmango:
This should be added to the opengl side too in `/src/gl/gl_hud.cpp`

kkartaltepe:
> This should be added to the opengl side too in `/src/gl/gl_hud.cpp`

What did you have in mind? The change to main.cpp was copied from https://github.com/flightlessmango/MangoHud/blob/master/src/gl/gl_hud.cpp#L137 and I have tested this on opengl, but not vulkan assuming they behaved similarly.

flightlessmango:
Ah yes, my mistake. Looks good!

PR #902, opened 2023-01-18
Fix FPS limiter

### LINKED ISSUE #879: FPS limiter causes high variability in FPS
In the following image you can observe an average of 60 FPS, however frame times range from 12 ms to 22 ms. I am convinced this happens because of MangoHud's framerate limiter.

Screenshot from 2023-01-04 22-07-22

So this is basically how I understand the FPS limiter is implemented in MangoHud. For each frame the following steps are executed
1. Sleep for some time
2. Let the game do its thing (update state, draw calls, etc.)
3. Swap buffer

To get a consistent FPS you would need to predict how long the 2nd step will take, and based on that time, compute for how long to sleep on step 1. The current implementation 'predicts' that step 2 will take the same time it took during the previous frame.

You can see how this could cause undesired variability in FPS. Lets suppose your target frametime is 16.6 ms (60 FPS) and, during the previous frame, step 2 took 4 ms. Then the algorithm 'predicts' this frame's 2nd step will also take 4 ms, so it decides to sleep for 16.6-4=12.6 ms in step 1. But if this frame's 2nd step takes 8 ms, you get a frametime of 12.6+8=20.6 ms.

In my opinion, a better way to implement this would be the following:

1. Let the game do its thing (update state, draw calls, etc.)
2. Sleep some amount of time (this time can be computed exactly, no predictions required)
3. Swap buffer

A disadvantage of this second approach is you get slightly higher latency, but you solve the first issue.

The fix is easy, just swap the steps in the relevant functions (glXSwapBuffers, glXSwapBuffersMscOML, etc.). By doing that I got the following result.

Screenshot from 2023-01-04 22-11-39

azn1999:
Note: the egl limiter is already correct

Fixes #879

### DIFF
--- a/src/gl/inject_glx.cpp
+++ b/src/gl/inject_glx.cpp
@@ -156,14 +156,13 @@ EXPORT_C_(void) glXSwapBuffers(void* dpy, void* drawable) {
     glx.Load();
 
     do_imgui_swap(dpy, drawable);
-    glx.SwapBuffers(dpy, drawable);
-
     using namespace std::chrono_literals;
     if (!is_blacklisted() && fps_limit_stats.targetFrameTime > 0s){
         fps_limit_stats.frameStart = Clock::now();
         FpsLimiter(fps_limit_stats);
         fps_limit_stats.frameEnd = Clock::now();
     }
+    glx.SwapBuffers(dpy, drawable);
 }
 
 EXPORT_C_(int64_t) glXSwapBuffersMscOML(void* dpy, void* drawable, int64_t target_msc, int64_t divisor, int64_t remainder)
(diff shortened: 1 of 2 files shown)


azn1999:
PS: I don't know much about Vulkan, so it's possible what I did is incorrect, but it worked fine in Doom Eternal.

flightlessmango:
I would assume we want the same thing for inject_egl. can you add that as well?

azn1999:
I think it's currently correct, since (the original) eglSwapBuffers is already called immediately after the FpsLimiter function. The problem occurs when the swap happens immediately before FpsLimiter.

PR #905, opened 2023-01-18
Logging: Switch to a different sleep function

TheComputerGuy96:
This fixes a freeze with zink when automatic logging is enabled

I didn't test this on Windows, so please report any issues there


flightlessmango:
I don't think this will cause any issues as we use this method in many other places

PR #908, opened 2023-01-21
X11 keyboard poller thread

jackun:
Idk why but seems to work :P


flightlessmango:
no idea why we didn't have x11 key polling in a thread already, this is pretty great

Issue #868, opened 2022-12-18
CPU Temperature not working


TateFleming:
I have a Ryzen 9 7900X CPU. Lm-sensors does not report the CPU sensor stats, so this tool fails on it. Hopefully this will be supported in the future. In the meantime, I put in a work around (just look at nvme temperature instead of cpu) so the program still functions.

Another problem was the program looking for hwmon directories in the incorrect locations. I put in a quick-fix that works for my build, so I can now display full GPU stats.

If anyone runs into these issues and needs help, feel free to reach out. Better yet, if anyone has solutions to these problems, please let me know!

gort818:
For the CPU temp, upgrade your kernel.

You had a problem displaying GPU stats?

TateFleming:
Kernel is latest 5.15. No CPU stats reported from sensors, seems to be a known thing.

It was assuming GPU stats were at '/sys/class/drm/card0-HDMI-A-1/device/hwmon/' which returned an error. The stats are actually in '/sys/class/drm/card0-HDMI-A-1/device/device/hwmon/.' I don't know if this is something specific to my build or if it is new with Zen4. I'm new to all of this myself, so I could be doing something incorrectly.

gort818:
That kernel is over a year old, I believe the monitor for those cpus was added about 5 months to the kernel.

Those are harmless errors, we iterate over the files to find the right one

TateFleming:
Now kernel is 6.1.0. CPU temperature updates, but all GPU stats are messed up now.

gort818:
What GPU, what exactly is messed up?

TateFleming:
GPU is Radeon RX6600.

I'm trying to display CPU and GPU stats, temp, power. Also GPU core and mem clocks, and CPU mhz.

GPU showing 0%, cooler than expected temperature, constant 65535MHz for one of the clocks, and ~10W power. I have two amdgpu items that show up when 'sensors' is ran, it appears that MangoHud is grabbing the second when it should grabbing the first potentially?

CPU stats are good besides temperature is stuck at 655C.

VRAM displays 0.0GiB.

gort818:
are you on the latest build of MangoHud?

TateFleming:
Yes I am. I just removed it completely and cloned the repo again to verify.

gort818:
7900X has an igpu doesn't it? 

can you try this?
https://github.com/flightlessmango/MangoHud/issues/802#issuecomment-1186574641

[...]

PR #912, opened 2023-01-24
Fix for AMD Ryzen 7000 CPU when using the iGPU

sfrcorne:
partly fixes the issue: #868

This fixes the values reported by MangoHud for AMD Ryzen 7000 CPUs when using the iGPU, so without a dedicated/discrete GPU.

When using a dGPU, all values except for `cpu_power` should already report the correct value.

For the `cpu_power` when using a AMD Ryzen APU + AMD Radeon dGPU, we might be able to read this value from the `gpu_metrics` file of the APU but this would require us to read two separate `gpu_metrics` files. Code for this is not included in this PR.

### DIFF
--- a/src/amdgpu.cpp
+++ b/src/amdgpu.cpp
@@ -28,17 +28,21 @@ bool amdgpu_verify_metrics(const std::string& path)
         return false;
     }
 
-    switch (header.format_revision)
-    {
-		case 1:
+	switch (header.format_revision)
+	{
+		case 1: // v1_1, v1_2, v1_3
+			if(header.content_revision<=0 || header.content_revision>3)// v1_0, not naturally aligned
+				break;
 			cpuStats.cpu_type = "GPU";
 			return true;
-		case 2:
+		case 2: // v2_1, v2_2, v2_3
+			if(header.content_revision<=0 || header.content_revision>3)// v2_0, not naturally aligned
+				break;
 			cpuStats.cpu_type = "APU";
 			return true;
-		default: 
+		default:
 			break;
-    }
+	}
 
     SPDLOG_WARN("Unsupported gpu_metrics version: {}.{}", header.format_revision, header.content_revision);
     return false;
@@ -47,7 +51,7 @@ bool amdgpu_verify_metrics(const std::string& path)
 #define MAX(x, y) (((x) > (y)) ? (x) : (y))
 void amdgpu_get_instant_metrics(struct amdgpu_common_metrics *metrics) {
 	FILE *f;
-	void *buf[MAX(sizeof(struct gpu_metrics_v1_3), sizeof(struct gpu_metrics_v2_3))];
+	void *buf[MAX(sizeof(struct gpu_metrics_v1_3), sizeof(struct gpu_metrics_v2_3))/sizeof(void*)+1];
 	struct metrics_table_header* header = (metrics_table_header*)buf;
 
 	f = fopen(metrics_path.c_str(), "rb");
@@ -55,8 +59,8 @@ void amdgpu_get_instant_metrics(struct amdgpu_common_metrics *metrics) {
 		return;
 
 	// Read the whole file
-	if (!fread(buf, sizeof(buf), 1, f) == 0) {
-		SPDLOG_DEBUG("Failed to read amdgpu metrics file '{}'", metrics_path.c_str());
+	if (fread(buf, sizeof(buf), 1, f) != 0) {
+		SPDLOG_DEBUG("amdgpu metrics file '{}' is larger than the buffer", metrics_path.c_str());
 		fclose(f);
 		return;
 	}
@@ -83,6 +87,12 @@ void amdgpu_get_instant_metrics(struct amdgpu_common_metrics *metrics) {
 
 		metrics->average_gfx_power_w = amdgpu_metrics->average_gfx_power / 1000.f;
 		metrics->average_cpu_power_w = amdgpu_metrics->average_cpu_power / 1000.f;
+		// Use cpu + soc power similar to 'get_cpu_power_k10temp' in 'cpu.cpp' ?
+		//metrics->average_cpu_power_w = amdgpu_metrics->average_cpu_power / 1000.f + amdgpu_metrics->average_soc_power / 1000.f;
+
+		// average_cpu_power_w fallback ?
+		//if(!( amdgpu_metrics->average_cpu_power ^ 0xffff && amdgpu_metrics->average_soc_power ^ 0xffff ))
+		//    metrics->average_cpu_power_w = amdgpu_metrics->average_socket_power / 1000.f - amdgpu_metrics->average_gfx_power / 1000.f;
 
 		metrics->current_gfxclk_mhz = amdgpu_metrics->current_gfxclk;
 		metrics->current_uclk_mhz = amdgpu_metrics->current_uclk;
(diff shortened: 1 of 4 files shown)


src/amdgpu.cpp
```diff
@@ -24,19 +24,21 @@ bool amdgpu_verify_metrics(const std::string& path)
...
-    }
+	switch (header.format_revision)
+	{
+		case 1: // v1_1, v1_2, v1_3
+			if(header.content_revision<=0 || header.content_revision>3)// v1_0, not naturally aligned
+				break;
+			cpuStats.cpu_type = "GPU";
+			return true;
+		case 2: // v2_1, v2_2, v2_3
+			if(header.content_revision<=0 || header.content_revision>3)// v2_0, not naturally aligned
```
jackun, line 35: > not naturally aligned

copy/paste error?

sfrcorne: according to the source code:

https://elixir.bootlin.com/linux/latest/source/drivers/gpu/drm/amd/include/kgd_pp_interface.h#L672

both v1_0 and v2_0 are not naturally aligned and not recommended.

flightlessmango:
Again I'm not convinced this shouldn't be fixed upstream.
We don't want to work around things that are going to change anyway

sfrcorne:
I hope you are right about the fact that they will fix this in the kernel. I will make a bug report this weekend on:
https://gitlab.freedesktop.org/drm/amd/-/issues
And try to get this fixed in the kernel. Looking at the git history of the relevant file, it seems to me that this was intentional (not updating all the data in `gpu_metrics` that is). But I do hope I am wrong about this.

Once they fix it in the kernel, then this PR is indeed not necessary any more and can be closed.

sfrcorne:
I have asked around in the mail list of https://gitlab.freedesktop.org/drm/amd and got the following answer from `Alex Deucher`:

> The metrics provided by the SMU firmware varies from asic to asic.

> You can see what metrics are supported for your asic in smu_v13_0_5_get_gpu_metrics() as that function populates the supported fields from the firmware to the common structure.  current_gfxclk is not supported in your asic, but average_gfxclk_frequency is.  So you'd want to use whichever field is available for a particular asic in Mangohud.

You can find the whole mail response here: https://lists.freedesktop.org/archives/amd-gfx/2023-February/089323.html

Can you tell me what your opinion on this is?

flightlessmango:
Yeah that's enough for me, we do indeed need a work around in this case

jackun:
Hmm, why are you merging master into this? Mess the branch up.

sfrcorne:
Ah sorry, I was using this branch for building my own custom package. I didn't know this would mess things up. I am not that familiar with working with git. I will look into the `rebase` command.

Is there still a way to fix this? I can always make a new branch & pull request if that works better for you?

jackun:
I'm assuming your `master` is following flightlessmango's then just `git rebase master`. Then you'll probably currently get a merge conflict. Mostly the prototype defines in `src/amdgpu.h`. Fix these, check it compiles and works, mark it resolved with `git add src/amdgpu.h && git rebase --continue` and then just keep force push this branch, when necessary.

sfrcorne:
oké, thanks. I think I fixed it.

gort818:
I recently got a 7900X will test this out soon.

gort818:
Ok seems to work fine, thanks for this. I am not a fan of using `average_gfxclk_frequency` as the output is not as nice as `current_gfxclk` but looks like at the moment we don't have another option.

sfrcorne:
If `current_gfxclk` is available, then it will still use `current_gfxclk`. Only when `current_gfxclk` is not provided will it try to use `average_gfxclk_frequency` as a fallback value. So for most people, it should still use `current_gfxclk`.

gort818:
Yeah I just meant for amd zen4

src/amdgpu.cpp
```diff
@@ -88,23 +89,87 @@ void amdgpu_get_instant_metrics(struct amdgpu_common_metrics *metrics) {
...
-		metrics->average_cpu_power_w = amdgpu_metrics->average_cpu_power / 1000.f;
-		// Use cpu + soc power similar to 'get_cpu_power_k10temp' in 'cpu.cpp' ?
-		//metrics->average_cpu_power_w = amdgpu_metrics->average_cpu_power / 1000.f + amdgpu_metrics->average_soc_power / 1000.f;
 
-		// average_cpu_power_w fallback ?
-		//if(!( amdgpu_metrics->average_cpu_power ^ 0xffff && amdgpu_metrics->average_soc_power ^ 0xffff ))
-		//    metrics->average_cpu_power_w = amdgpu_metrics->average_socket_power / 1000.f - amdgpu_metrics->average_gfx_power / 1000.f;
+		if( IS_VALID_METRIC(amdgpu_metrics->average_cpu_power) && IS_VALID_METRIC(amdgpu_metrics->average_soc_power) ) {
+			// Use 'cpu power' + 'soc power' similar to 'get_cpu_power_k10temp' in 'cpu.cpp'
+			metrics->average_cpu_power_w = amdgpu_metrics->average_cpu_power / 1000.f + amdgpu_metrics->average_soc_power / 1000.f;
```
flightlessmango, line 95: Why would cpu_power be cpu_power + soc_power ?

sfrcorne: From how I understand it `cpu_power` is the cpu core power while `soc_power` is the soc power. To make the reported power more consistent with what is reported by the function `get_cpu_power_k10temp` in the file `cpu.cpp`, I used cpu core power + soc power.

The function `get_cpu_power_k10temp` in the file `cpu.cpp` uses the following code:
```
power = (coreVoltage * coreCurrent + socVoltage * socCurrent) / 1000000;
```

I tried to look up the meaning of `cpu_power` again just now but could not find additional information on it. I don't know anymore where I got the information/idea that `cpu_power` is only the cpu core power, so I could be wrong. If you don't like it, then I can always change it to only use `cpu_power`.

sfrcorne: Sorry for the late replay, the comment was "PENDING" and I didn't notice it.

sfrcorne: Do you want me to change the code to the following?

```
metrics->average_cpu_power_w = amdgpu_metrics->average_cpu_power / 1000.f;
```

flightlessmango: Yes I think that should be better

src/amdgpu.cpp
```diff
@@ -49,6 +49,7 @@ bool amdgpu_verify_metrics(const std::string& path)
 }
 
 #define MAX(x, y) (((x) > (y)) ? (x) : (y))
+#define IS_VALID_METRIC(FIELD) (FIELD != 0xffff)
```
flightlessmango, line 52: Is there another method we could determine if it's valid or not?
My concern is that it can be valid and still be 0

sfrcorne: The last mail I received from the freedesktop mail list was the following:
```
[AMD Official Use Only - General]

For some members, "0" is a valid value. 
Thus "0xffff" is used instead to tell the output is invalid/unsupported.

BR
Evan
```
So I think that this is the best method for determining if it is valid or not.

flightlessmango: Yeah sounds great then :+1:

sfrcorne: Here is a link to that email: https://lists.freedesktop.org/archives/amd-gfx/2023-February/089461.html

In case you are interested.

src/amdgpu.cpp
sfrcorne: oké, I will change it then.

PR #922, opened 2023-02-10
blacklist: add SteamVR's vrcompositor

Atemu:
Limiting frame rate in vrcompositor has catastrohpic consequences and it's not something the user will ever see the HUD on. Every SteamVR user with a global frame rate limit has to explicitly blacklist vrcompositor, so it's better to have this upstream.


Atemu:
Ping @flightlessmango

Atemu:
Thanks!

PR #925, opened 2023-02-15
Use libdrm_amdgpu as an alternative GPU load information source

### LINKED ISSUE #923: GPU utilization always at 100% on Vega 3
GPU utilization always shows 100% even when idling. I'm sure it isn't MangoHud's fault, because `/sys/class/drm/card0/device/gpu_busy_percent` is `100` all the time too. [I've seen this issue already mentioned here](https://gitlab.freedesktop.org/drm/amd/-/issues/1932) 10 months ago, with no reply.

What's interesting is that [radeontop](https://github.com/clbr/radeontop) seems to display the GPU utilization correctly. After taking a look at the source code, I've concluded that it calculates the percentage by repeatedly sampling `amdgpu_read_mm_registers` (take a look at amdgpu.c:25), &-ing it with `1U<<31` (ticks.c:53) and then dividing the number of times it wasn't 0 by the number of samples (ui.c:141). Can such a mechanism be implemented in MangoHud as a workaround for some problematic Vega GPUs?

bsolos:
Use libdrm_amdgpu to calculate the GPU load. This should resolve #923.

The GPU load calculation method was inspired by [radeontop](https://github.com/clbr/radeontop), but no code in this PR was copied from there.


RPINerd (approved):
I don't have a vega card to physically test this fork, but the general structure looks good and I didn't find anything basic like spelling errors 😃

bsolos:
It works properly on my machine, and I don't have any other Vega cards to test it on. Actually, it should also work on most non-Vega AMD GPUs

jackun:
https://github.com/flightlessmango/MangoHud/pull/703

Also you'd need drm master authorization, with secondary GPUs at least.

meson.build
```diff
@@ -88,10 +88,12 @@ if is_unixy
   dep_wayland_client = dependency('wayland-client',
       required: get_option('with_wayland'), version : '>=1.11')
   dbus_dep = dependency('dbus-1', required: get_option('with_dbus')).partial_dependency(compile_args : true, includes : true)
+  libdrm_amdgpu_dep = cpp.find_library('drm_amdgpu', required: get_option('with_libdrm_sampling'))
```
evelikov, line 91: s/cpp.find_library/dependency/

src/amdgpu.cpp
```diff
@@ -171,7 +172,14 @@ void amdgpu_get_metrics(){
...
 	amdgpu_common_metrics_m.lock();
-	gpu_info.load = amdgpu_common_metrics.gpu_load_percent;
+
+#ifdef HAVE_LIBDRM_SAMPLING
+	if (!do_libdrm_sampling) {
+#endif
+		gpu_info.load = amdgpu_common_metrics.gpu_load_percent;
+#ifdef HAVE_LIBDRM_SAMPLING
+	}
+#endif
```
evelikov, line 182: The value gets overwritten, so the above change should be needed. Right?

src/amdgpu_libdrm.cpp
```diff
@@ -0,0 +1,85 @@
+#include <deque>
+#include <mutex>
+#include <thread>
+#include <chrono>
+#include <libdrm/amdgpu.h>
```
evelikov, line 5: `s|libdrm/||` - the `dependency` suggestion above will ensure the correct folder is passed during compilation.

src/amdgpu_libdrm.h
```diff
@@ -0,0 +1,26 @@
...
+    int busy;
+};
+
+extern std::string dri_device_path;
+extern bool do_libdrm_sampling;
+
+void libdrm_do_sample();
+void libdrm_thread();
+void libdrm_get_info();
+int libdrm_initialize();
```
evelikov, line 26: Of the 4 above, only `libdrm_get_info()` should remain here. The rest ought to be annotated as static within the .C file.

Might need to move `libdrm_get_info` within the bottom of the (C) file to avoid forward declarations.

src/gpu.cpp
```diff
@@ -74,14 +75,20 @@ nvapi_util();
...
     if (metrics_path.empty()){
-        if (amdgpu.busy) {
-            rewind(amdgpu.busy);
-            fflush(amdgpu.busy);
-            int value = 0;
-            if (fscanf(amdgpu.busy, "%d", &value) != 1)
-                value = 0;
-            gpu_info.load = value;
+#ifdef HAVE_LIBDRM_SAMPLING
+        if (!do_libdrm_sampling) {
```
evelikov, line 79: Perhaps it's worth making this global always 0, when `HAVE_LIBDRM_SAMPLING` is not defined? This way one can drop the repeated `ifdef` guards.

evelikov: Actually: it might even better to unconditionally build with it. The library is guaranteed to be present on literally all Linux/FreeBSD/Unixy distro that wants GL or Vulkan.

In addition, as-is it's possible to build w/o that then get hella confused why the toggle does nothing.

evelikov:
> Also you'd need drm master authorization, with secondary GPUs at least.

This should not be needed, if one uses the renderD node - seems like currently the card one is being used.

If the card node is already opened you can the fd with `drmGetRenderDeviceNameFromFd()`. Alternatively `drmGetDevices2()`  gives you all devices, find the needed one by matching the card node and use the render node. [This MR](https://github.com/intel/libva-utils/pull/292) does something different but should give you a good starting point.

evelikov:
@bsolos overall I would encourage you to open a bug at the [AMDGPU gitlab](https://gitlab.freedesktop.org/drm/amd/-/issues), clearly describe the issue (exact kernel, version, distro, use-case, etc) + CC Alex Deucher aka @agd5f

This is a great workaround, but the upstream gpu metrics should really be fixed.

bsolos:
I didn't open an issue there because https://gitlab.freedesktop.org/drm/amd/-/issues/1932 is open already. Should I still open a new one?

evelikov:
One should not assume that devs don't care about issues, just because there's no update. Sometimes they have higher/other priorities, sometimes it fall through the cracks.

By opening/prodding you'll increase visibility and raise severity. If you can test kernel patches, it's more likely that devs will try to get fixed faster. Sitting quietly does not help, I'm afraid.

src/amdgpu_libdrm.cpp
bsolos: I think that `libdrm/` can't be omitted here, even with `dependency` instead of `cpp.find_library`

evelikov: If that doesn't work either the dependency is not added to the correct place in meson.build or the distro is horribly broken.

bsolos: I might need to add `libdrm` as a separate dependency for this to work. I'll try it shortly

bsolos: It seems that the problem is that it includes the local `./amdgpu.h` instead of the global `/usr/include/libdrm/amdgpu.h` and I'm not sure what to do about that

src/amdgpu.cpp
bsolos: It seems like I've made an incorrect assumption that it should get overwritten, but it actually doesn't. Will remove in the next commit

src/amdgpu_libdrm.cpp
```diff
@@ -51,6 +52,13 @@ static int libdrm_initialize() {
         return -1;
     }
 
+    char *renderD = drmGetRenderDeviceNameFromFd(fd);
+    fd = open(renderD, O_RDWR);
```
evelikov, line 56: Err, please use the DRMDevice API, don't open one node only to get another.

bsolos: Sorry, I've never really worked with libdrm before. Will fix shortly

agd5f:
Wouldn't something like:
https://www.kernel.org/doc/html/latest/gpu/drm-usage-stats.html
make more sense then polling hardware registers?  Plus it's cross-vendor.

Alex

bsolos:
> @bsolos overall I would encourage you to open a bug at the [AMDGPU gitlab](https://gitlab.freedesktop.org/drm/amd/-/issues), clearly describe the issue (exact kernel, version, distro, use-case, etc) + CC Alex Deucher aka @agd5f
> 
> This is a great workaround, but the upstream gpu metrics should really be fixed.

[Seems like the load sensor isn't supported on the hardware level](https://gitlab.freedesktop.org/drm/amd/-/issues/1932)

src/meson.build
```diff
@@ -89,6 +89,7 @@ if is_unixy
     'battery.cpp',
     'control.cpp',
     'gamepad.cpp',
+    'amdgpu_libdrm.cpp'
```
evelikov, line 92: Nit: add a trailing comma here + in the dependency arrays further down.

src/amdgpu_libdrm.h
```diff
@@ -0,0 +1,25 @@
...
+};
+
+struct libdrm_sample {
+    bool busy_bit;
+};
+
+struct libdrm_stats {
+    int busy;
+};
+
```
evelikov, line 21: All the above information is private to amdgpu_libdrm.cpp, so it should be moved there.

src/amdgpu_libdrm.cpp
```diff
@@ -0,0 +1,112 @@
...
+    drmDevicePtr devices[LIBDRM_MAX_DEVICES];
+    int device_count = drmGetDevices2(0, devices, LIBDRM_MAX_DEVICES);
+    if (device_count < 0) {
+        SPDLOG_ERROR("drmGetDevices2 failed");       
+        return -1;
+    }
+
+    char *renderd_node = nullptr;
+    for (int i = 0; i < device_count; i++) {
+        if ((devices[i]->available_nodes & 0b101) != 0b101) {
```
evelikov, line 58: Using `DRM_NODE_PRIMARY | DRM_NODE_RENDER` over the binary numbers, makes this easier to read.

src/amdgpu_libdrm.cpp
```diff
@@ -0,0 +1,112 @@
...
+            continue;
+        }
+
+        if (devices[i]->nodes[DRM_NODE_PRIMARY] == dri_device_path) {
```
evelikov, line 62: Disclaimer C++ is rusty: String comparison usually happens with `strcmp()`. This will compare the pointers.

src/amdgpu_libdrm.cpp
```diff
@@ -0,0 +1,112 @@
...
+            renderd_node = devices[i]->nodes[DRM_NODE_RENDER];
+            break;
+        }
+    }
+
+    if (renderd_node == nullptr) {
+        SPDLOG_ERROR("No renderD node found for '{}'", dri_device_path);
+        return -1;
```
evelikov, line 70: Missing `drmFreeDevices()` before all the return statements?

src/amdgpu_libdrm.cpp
evelikov: Not sure what you mean here - there is no `amdgpu.h` in tree to get confused with. Unless there is a local one in your git checkout?

src/amdgpu_libdrm.cpp
bsolos: `dri_device_path` is an std::string and not a char*. Would that still compare just the pointers?

meson.build
```diff
@@ -88,10 +88,14 @@ if is_unixy
   dep_wayland_client = dependency('wayland-client',
       required: get_option('with_wayland'), version : '>=1.11')
   dbus_dep = dependency('dbus-1', required: get_option('with_dbus')).partial_dependency(compile_args : true, includes : true)
+  libdrm_dep = dependency('libdrm', required: true)
+  libdrm_amdgpu_dep = dependency('libdrm_amdgpu', required: true)
```
evelikov, line 92: Nit: drop the `required: true` - it's the default.

Also it was me, I would squash all the commits into one. As-is things are quite messy.

HTH o/

src/amdgpu_libdrm.cpp
bsolos: [There absolutely is an amdgpu.h in tree](https://github.com/flightlessmango/MangoHud/blob/master/src/amdgpu.h)

src/amdgpu_libdrm.cpp
evelikov: Right, didn't see that. AFAICT there should be a `bool operator== (const std::string& c1, const char *)` in the system headers, making this work correctly. Ignore my comment.

src/amdgpu_libdrm.cpp
bsolos: `src/` seems to be on the include path for some reason

src/amdgpu_libdrm.cpp
bsolos: `DRM_NODE_PRIMARY | DRM_NODE_RENDER` results in `0b10` and not `0b101`. I couldn't find any way I could use that, maybe because I'm really bad at bitmasks

evelikov: Note to self - don't review before coffee is drunk:

```c
#define THE_MASK (1 << DRM_NODE_PRIMARY | 1 << DRM_NODE_RENDER)

if ((devices[i]->available_nodes & THE_MASK != THE_MASK ) {
```

src/amdgpu_libdrm.cpp
evelikov: > src/ seems to be on the include path for some reason

This should make no different what-so-ever. What is the exact `#include` directive you're using and the error log being produced?

src/amdgpu_libdrm.cpp
bsolos: Thanks

src/amdgpu_libdrm.cpp
bsolos: With `#include <amdgpu.h>` it outputs
```
../src/amdgpu_libdrm.cpp:39:1: error: ‘amdgpu_device_handle’ does not name a type
   39 | amdgpu_device_handle amdgpu_handle;
      | ^~~~~~~~~~~~~~~~~~~~
../src/amdgpu_libdrm.cpp: In function ‘void libdrm_do_sample(libdrm_sample*)’:
../src/amdgpu_libdrm.cpp:43:30: error: ‘amdgpu_handle’ was not declared in this scope
   43 |     amdgpu_read_mm_registers(amdgpu_handle, LIBDRM_GRBM_STATUS / 4, 1, 0xffffffff, 0, &registers);
      |                              ^~~~~~~~~~~~~
../src/amdgpu_libdrm.cpp:43:5: error: ‘amdgpu_read_mm_registers’ was not declared in this scope
   43 |     amdgpu_read_mm_registers(amdgpu_handle, LIBDRM_GRBM_STATUS / 4, 1, 0xffffffff, 0, &registers);
      |     ^~~~~~~~~~~~~~~~~~~~~~~~
../src/amdgpu_libdrm.cpp: In function ‘int libdrm_initialize()’:
../src/amdgpu_libdrm.cpp:103:69: error: ‘amdgpu_handle’ was not declared in this scope
  103 |     if (amdgpu_device_initialize(fd, &libdrm_major, &libdrm_minor, &amdgpu_handle)) {
      |                                                                     ^~~~~~~~~~~~~
../src/amdgpu_libdrm.cpp:103:9: error: ‘amdgpu_device_initialize’ was not declared in this scope
  103 |     if (amdgpu_device_initialize(fd, &libdrm_major, &libdrm_minor, &amdgpu_handle)) {
      |         ^~~~~~~~~~~~~~~~~~~~~~~~
```

If I add an `#include <libdrm/amdgpu.h>` to the local `./amdgpu.h` it compiles. IMO the only possible explanation for this is that the compiler includes the local one, even with the angle brackets

evelikov:
```
c++ -Isrc/libMangoHud.a.p -Isrc -I../src -I../include -Isubprojects/imgui-1.81 -I../subprojects/imgui-1.81 ... -I/usr/include/libdrm ... -c ../src/amdgpu_libdrm.cpp
../src/amdgpu_libdrm.cpp:39:1: error: ‘amdgpu_device_handle’ does not name a type
```

From the above - meson is inserting `-Isrc -I../src` at the very start, even though there is no `include_directories` directives instructing it to. Perhaps something in our meson.build is forcing it or maybe not. In the latter case this is fairly concerning.

As a quick and dirty approach I would simply rename the in-tree file - as a separate commit.

bsolos: Meson seems to always -I the directory in which the source files reside. Just tested it on an almost empty project

agd5f:
The other problem with polling registers is that it keeps the GPU awake
using more power.  The driver has to disable gfxoff when you read back
registers.

Alex

bsolos:
`drmGetStats` doesnt seem to work on the renderD node, and using it on the primary node always gives stats.count=0. Maybe I have to authenticate first?

evelikov:
@agd5f perhaps a in-tree kernel AMDGPU doc outlining the preferred options and their caveats will be great. Something people can keep an eye on, as things evolve - say team introduces new method do fetch X, or method Y has issues (aka gfxoff issue mentioned), approach Z might be deprecated (ETA, reason), etc.

evelikov:
@bsolos `drmGetStats` is legacy API and should not be used. As the in-kernel comment says "getstats is defunct, just clear"

bsolos:
This makes sense now. It seems like finding what is the correct way is much more difficult than I thought

agd5f:
For reference on how to use the fdinfo interface see:
https://www.spinics.net/lists/intel-gfx/msg294401.html

evelikov:
@bsolos the site/link is down see https://patchwork.freedesktop.org/series/102175/

@agd5f does that interface provide system-wise statistics? it seems to be per-client and per-fd, where mangohud exposes the total system data. Technically one could iterating over `/proc/foo/fdinfo` for the total, assuming they have permissions - yet mangohud should not be run as root.

agd5f:
Yes, it's per client.  Similar to top for the CPU.

Alex

Umio-Yasuno:
Hello.
I am developing [amdgpu_top](https://github.com/Umio-Yasuno/amdgpu_top).  
`amdgpu_top` has simple `fdinfo` parser and performance counters (GRBM, GRBM2, CP_STAT) readings and sensor readings implemented.  
Would it help you?

However, to use it like `intel_gpu_top` requires a change.

Umio-Yasuno:
working branch: https://github.com/Umio-Yasuno/amdgpu_top/tree/json-output

PR #929, opened 2023-02-17
Add emphasis about using git- AUR installs

RPINerd:
Cleaned up lots of markdown best practices, organized some and updated the table of contents


flightlessmango:
Seems like overkill to add `bash` to every command, especially to shell scripts that already define it. Otherwise it's looking good 👍

RPINerd:
I find it to be overkill too from a visual perspective looking at the raw markdown haha, but apparently it's "bad practice" to not specify a language after the ```? I'll cut all the extraneous bash specifiers out and make another commit :)

Issue #920, opened 2023-02-03
Add option to toggle between old and new FPS limiter


hannesmann:
I use the FPS limiter in MangoHud on a FreeSync monitor to get rid of tearing while avoiding the additional latency of V-sync and #902 breaks that behavior. It's easy to tell the latency is worse since it behaves the same as V-sync now ("wait and present frame" instead of "present frame and wait").

I'm not seeing the variability in frametimes that was mentioned in the issue. Minecraft is slightly unstable but this happens regardless if the old limiter is active or not, the game just seems to behave that way ([off](https://i.imgur.com/3fMmyc5.png) vs [on](https://i.imgur.com/0Auiwae.png)). Other games like WoW are very stable, usually staying in a range of 7.1-7.2 ms when limited to 140 FPS.

The other two FPS limiters that I know of on Linux, libstrangle and DXVK, both wait *after* presenting the frame:
https://gitlab.com/torkel104/libstrangle/-/blob/ed03e3f6/src/opengl/glx.c#L78
https://github.com/doitsujin/dxvk/blob/421d7d90/src/vulkan/vulkan_presenter.cpp#L82

IMO the old behavior should be restored as the default with an option to change how the limiter works (something like `fps_limit_method`?).

hannesmann:
Forgot to mention that this is on AMD (RX 6700 XT). Maybe it behaves differently on NVIDIA.

flightlessmango:
If it increases input lag, then I do agree it should be reverted.
I'm gonna leave this open for a bit in case people have some arguments as to why it should remain

KajzerD:
Can confirm, new fps limiter does cause some sort of input lag. It's easily noticable in Apex Legends, Hunt Showdown and other competitive titles. I have tried comparing with it ON and OFF while in game, definitely noticable for competitive players like me. Having an option to switch between fps limiters would be great. DXVK and Libstrangle don't cause this issue

hannesmann:
@KajzerD Could you try #932 and confirm it restores the old behavior?

KajzerD:
@hannesmann I will try it once I come home from work. Do I need to compile it manually from git or? Since I'm not good with that stuff. I always relied on mangohud sh script to install stuff. And which options would I be using specifically in config to make this happen?

PR #932, opened 2023-02-20
Add "FPS limit method" parameter

hannesmann:
This adds an option to select if the FPS limiter should wait before (`early`) or after (`late`) presenting a frame. The default is set to `late` which matches MangoHud 0.6.8, DXVK and libstrangle.

Waiting before present is required for 100% stable frametimes (since the limiter can know exactly how much time is left in every frame) but it adds latency (since the limiter adds a delay before sending the frame to the display). 

Fixes: #920


flightlessmango:
I think we can merge this, can you just resolve the conflicts?

hannesmann:
Should be resolved now.

Issue #927, opened 2023-02-16
-Duse_system_vulkan=enabled fails with Vulkan-Headers >= 1.3.241


jbeich:
[gen_enum_to_str.py](https://github.com/flightlessmango/MangoHud/blob/master/bin/gen_enum_to_str.py) ([upstream version](https://github.com/Mesa3D/mesa/blob/main/src/vulkan/util/gen_enum_to_str.py)) currently lacks [mesa!21225](https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/21225)
```c++
In file included from vk_enum_to_str.c:28:
./vk_enum_to_str.h:329:39: error: unknown type name 'VkFaultLevel'
    const char * vk_FaultLevel_to_str(VkFaultLevel input);
                                      ^
./vk_enum_to_str.h:330:47: error: unknown type name 'VkFaultQueryBehavior'
    const char * vk_FaultQueryBehavior_to_str(VkFaultQueryBehavior input);
                                              ^
./vk_enum_to_str.h:331:38: error: unknown type name 'VkFaultType'
    const char * vk_FaultType_to_str(VkFaultType input);
                                     ^
./vk_enum_to_str.h:368:59: error: unknown type name 'VkPipelineCacheValidationVersion'; did you mean 'VkPipelineCacheHeaderVersion'?
    const char * vk_PipelineCacheValidationVersion_to_str(VkPipelineCacheValidationVersion input);
                                                          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
                                                          VkPipelineCacheHeaderVersion
/usr/include/vulkan/vulkan_core.h:1189:3: note: 'VkPipelineCacheHeaderVersion' declared here
} VkPipelineCacheHeaderVersion;
  ^
In file included from vk_enum_to_str.c:28:
./vk_enum_to_str.h:370:49: error: unknown type name 'VkPipelineMatchControl'
    const char * vk_PipelineMatchControl_to_str(VkPipelineMatchControl input);
                                                ^
./vk_enum_to_str.h:391:48: error: unknown type name 'VkSciSyncClientTypeNV'
    const char * vk_SciSyncClientTypeNV_to_str(VkSciSyncClientTypeNV input);
                                               ^
./vk_enum_to_str.h:392:51: error: unknown type name 'VkSciSyncPrimitiveTypeNV'
[...]

PR #938, opened 2023-03-04
Fix the tests and add push/pr CI setup

evelikov:
Nothing too controversial there I believe, but shout if you thing anything needs more polish.

The biggest piece that comes off the top of my head is more CI testing. Namely, there are various references about Windows across the build/code, although it's not clear:
 - what is the magical `meson -Dfoo=disabled -Dbar=enabled` combination that should be used
 - are both MSVC and MINGW supported - which versions
 
 If anyone has pointers, I'm happy to wire those up as well

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
--- a/.github/workflows/build-package.yml
+++ b/.github/workflows/build-package.yml
@@ -18,7 +18,7 @@ jobs:
         run: |
           sudo apt update
           sudo apt install gcc-multilib g++-multilib ninja-build python3-setuptools python3-wheel mesa-common-dev libxnvctrl-dev libdbus-1-dev unzip
-          sudo pip3 install 'meson>=0.54' mako
+          sudo pip3 install 'meson>=0.60' mako
           wget https://github.com/KhronosGroup/glslang/releases/download/SDK-candidate-26-Jul-2020/glslang-master-linux-Release.zip
           unzip glslang-master-linux-Release.zip bin/glslangValidator
           sudo install -m755 bin/glslangValidator /usr/local/bin/
--- a/.github/workflows/build-package.yml
+++ b/.github/workflows/build-package.yml
@@ -10,8 +10,7 @@ on:
 
 jobs:
   build:
-    #runs-on: ubuntu-latest
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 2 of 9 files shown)


evelikov:
Forgot the Vulkan fix - now CI should be a bit happier

evelikov:
Closes: https://github.com/flightlessmango/MangoHud/issues/927

@jbeich can you give this a try? It should resolve the Vulkan build issue you're seeing

jbeich:
Sorry, I can't test as MangoHud doesn't build without [patches](https://github.com/freebsd/freebsd-ports/tree/main/benchmarks/mangohud/files) on FreeBSD. I don't even use MangoHud, so only filed #927 to help @danfe who maintains the package.

Looking at the commits here I don't see anything that would fix #927. `use_system_vulkan` option is `disabled` by default, and Ubuntu 22.04 (on CI) has Vulkan 1.3.204 but the affected version is 1.3.241 (37 versions ahead).

jbeich:
Note, #927 explicitly mentions `-Duse_system_vulkan=enabled` thus you cannot use `Closes #927` unless you drop `use_system_vulkan` entirely.

evelikov:
@jbeich [meson: do not pick system vulkan, when explicitly disabled](https://github.com/flightlessmango/MangoHud/pull/938/commits/cf5f0a84cad951269c1d8d037265d090902f09e0) fixes the exact same issues as the ones reported. Not sure what you meant with `... thus you cannot use Closes #927 unless you drop use_system_vulkan...`.

If you don't have the time to test this that's fine -  perhaps @danfe could?

jbeich:
> the exact same issues as the ones reported

The error maybe the same but the cause is different. Downstream may explicitly pass `-Duse_system_vulkan=enabled` (like FreeBSD), so would still be affected. A proper fix would be something like https://github.com/haasn/libplacebo/commit/baec0e53c959 MangoHud needs to fix its `vk.xml` parser, similar to libplacebo.

jbeich:
Err, s/libplacebo/Mesa/. MangoHud appears to use an old version of  `vk.xml` parser from Mesa.

- https://github.com/flightlessmango/MangoHud/blob/master/bin/gen_enum_to_str.py
- https://github.com/Mesa3D/mesa/blob/main/src/vulkan/util/gen_enum_to_str.py

evelikov:
I get it, you're not interested in testing sorry for troubling you :bow:

danfe:
> If you don't have the time to test this that's fine - perhaps @danfe could?

I will be updating FreeBSD port hopefully soonish™ and will try not to forget to take a closer look at this :) thanks for the heads-up!

evelikov:
@flightlessmango your earlier commit d0f8c8aa9498b2feeb1e13f5cd8638cf8a4d6594 broke mangocontrol and mangoapp - perhaps you forgot to `git add`? Can we have this PR merged, so such mistakes are caught earlier

evelikov:
> I have built mangohud with both in the past, but it's going to be borked because there will be unix specific things that need to be ifdef'd.
As for what version, I'm not sure. I've just used latest MINGW in arch

Thanks for the reference points. Will see if I can hook that up as follow-up.

PR #939, opened 2023-03-04
Drop unused warnings, add new one, parallelise

evelikov:
This PR drops a few no longer relevant warnings - `Wno-format-truncation` and the no-override ones. Adds a new one - `missing-declarations` - which kindly enough prompts about some dead code, which was removed.

Last by not least the meson `get_supported-arguments()` is used, which allows meson to do the checks in parallel, plus is aware of the `-Wno` trick and uses it as applicable.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -151,13 +151,6 @@ foreach a : ['non-virtual-dtor', 'missing-field-initializers', 'format-truncatio
   endif
 endforeach
 
-no_override_init_args = []
-foreach a : ['override-init', 'initializer-overrides']
-  if cc.has_argument('-W' + a)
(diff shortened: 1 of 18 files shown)


eli-schwartz:
> Last by not least the meson `get_supported-arguments()` is used, which allows meson to do the checks in parallel, plus is aware of the `-Wno` trick and uses it as applicable.

For the record, Meson doesn't do checks in parallel -- but a couple attempts have been made at fleshing out support for this, and a common theme is that we'd only be able to do it for functions like get_supported_arguments that take multiple things to check. So, this change is good and allows Meson to silently optimize it in the future. :)

More important as of now, it looks a lot prettier to use it.

PR #941, opened 2023-03-04
meson: remove mangoapp_32bit option

evelikov:
Was introduced as a means to shorten the build cycles. With the static lib helper it should no longer be needed.


jackun:
wtf is github's problem

evelikov:
Give me a minute - will fixup locally and force-push.

evelikov:
Fixed

PR #962, opened 2023-03-21
Install MangoHud.x86.json in pkgbuild/PKGBUILD

Billli11:
When packaging `lib32-mangohud`,  the path `/usr/share` is removed recursively.

Which also include file `/usr/share/vulkan/implicit_layer.d/MangoHud.x86.json`, cause mangohud fail to run on 32-bit application (for example borderlands 2).

This commit add back the file after removing `/usr/share`.

### DIFF
--- a/pkgbuild/PKGBUILD
+++ b/pkgbuild/PKGBUILD
@@ -76,4 +76,5 @@ package_lib32-mangohud() {
   DESTDIR="${pkgdir}" ninja -C build32 install
   rm -rf "$pkgdir/usr/bin"
   rm -rf "$pkgdir/usr/share"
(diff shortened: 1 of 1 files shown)


flightlessmango:
I don't think 32bit needs it's own json manifest. Should work with just one and then the loader figures it out

Billli11:
The manifest is needed.

Both is point to a difference library_path

`diff MangoHud.x86_64.json MangoHud.x86.json`
   7c7  
   <       "library_path": "/usr/lib/libMangoHud.so",
   ---  
   >       "library_path": "/usr/lib32/libMangoHud.so",

No mangohud when `Tiny Tina's Assault on Dragon Keep: A Wonderlands One-shot Adventure` with out the x86 version of manifest.

PR #973, opened 2023-04-02
Fix position problem to support more handheld devices with screen rotated

MiZai2:
I wrote this to fix position problem for handheld devices with screen rotated.

### DIFF
--- a/src/app/main.cpp
+++ b/src/app/main.cpp
@@ -205,9 +205,11 @@ static void msg_read_thread(){
 }
 
 static const char *GamescopeOverlayProperty = "GAMESCOPE_EXTERNAL_OVERLAY";
+static int screen_width  = 1280;
+static int screen_height = 800;
 
 static GLFWwindow* init(const char* glsl_version){
-    GLFWwindow *window = glfwCreateWindow(1280, 800, "mangoapp overlay window", NULL, NULL);
+    GLFWwindow *window = glfwCreateWindow(screen_width, screen_height, "mangoapp overlay window", NULL, NULL);
     Display *x11_display = glfwGetX11Display();
     Window x11_window = glfwGetX11Window(window);
     if (x11_window && x11_display)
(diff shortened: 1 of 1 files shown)


src/app/main.cpp
```diff
@@ -289,12 +309,15 @@ int main(int, char**)
...
     sw_stats.deviceName = deviceName;
-    if (deviceName.find("Radeon") != std::string::npos
+    if (vendor.find("AMD") != std::string::npos
+    || deviceName.find("Radeon") != std::string::npos
     || deviceName.find("AMD") != std::string::npos){
         vendorID = 0x1002;
-    } else if (deviceName.find("Intel") != std::string::npos) {
+    } else if (vendor.find("Intel") != std::string::npos
+    || deviceName.find("Intel") != std::string::npos) {
         vendorID = 0x8086;
```
evelikov, line 321: The vendor detection changes should probably be a separate PR.

src/app/main.cpp
```diff
@@ -245,11 +247,26 @@ static bool render(GLFWwindow* window) {
...
+    glfwSetWindowSize(window, screen_width, screen_height);
     ImGui::EndFrame();
     return last_window_size.x != window_size.x || last_window_size.y != window_size.y;
 }
 
+static void updateScreenInfo(){
+    GLFWmonitor* pMonitor = glfwGetPrimaryMonitor();
+    const GLFWvidmode * mode = glfwGetVideoMode(pMonitor);
+    screen_width = mode->width;
+    screen_height = mode->height;
```
evelikov, line 259: Will this work correctly when the primary (and only) monitor gets disconnected? Say when your dog pulls the cable for a moment, because they want you to play with them :dog2:

src/app/main.cpp
```diff
@@ -289,12 +309,15 @@ int main(int, char**)
     notifier.params = &params;
     start_notifier(notifier);
     window_size = ImVec2(params.width, params.height);
+    std::string vendor = (char*)glGetString(GL_VENDOR);
```
MiZai2:
```suggestion
```

MiZai2 (commented):
I wrote this to support more resolutions.
1.fix some problem cause by zero monitor since last commit.
2.recover vendor code.

src/app/main.cpp
MiZai2: I will update a new file to resolve those problems

PR #977, opened 2023-04-02
Add mangoplot script

AdelKS:
I removed my fork by mistake, forgetting about this PR, and I couldn't push to the old PR any longer, so I reopened another.

Hello guys, as I started benchmarking with `log_interval=0` to have accurate percentiles (MangoHud otherwise simply saves to disk 1 frame every log_interval), the produced files are too big to be handled by the website. A local way to plot and compare between benchmarks was lacking, so here it is.

This PR introduces a python script that enables comparing visually between benchmarks, and more accurately than what the website displays. It depends on `matplotlib` and needs a graphical back-end, this is left to the packagers to translate into their packaging system.

Here's an example output of the script (on a set of benchmarks I did)

benchmarks

It shows the distribution of the frame rates (`1 / (frame time)`), the 50%, the 1% and 0.1% percentiles. The `n`% percentiles are not exactly the `n`% lows: the `n`% lows involves averaging all the frame rates below the `n`% percentile.

Visualizing the distribution of the frame times enables seeing: 
- More nuances between benchmarks to properly assess what one prefers in one's own use-case
- That a benchmark is not reproducible yet, if e.g. a rebench doesn't visually have the same distribution

we can even switch to this view on the website, maybe, visualizing the frames is pretty, but pretty un-useful, except maybe to confirm that two benchmarks are related by seeing that they have the same trends.

On follow-up PRs, we can add some extra features like:
- Give which percentiles to display
- Accept a list of benchmark files instead of only a folder.
- Set an `fps_min` and an `fps_max` to crop the view (although it can be done with user input on the `matplotlib` window)

Feedback welcome !

### DIFF
--- a/bin/mangoplot.py
+++ b/bin/mangoplot.py
@@ -1,0 +1,424 @@
+#!/usr/bin/env python
+
+r"""
+    Script to plot all the MangoHud benchmarks contained in a given folder.
+"""
+from pathlib import Path
+import argparse
+import csv
+
+from typing import List, Union
+
+import numpy as np
+
+import matplotlib.pyplot as plt
+from matplotlib.widgets import Cursor
+
+from matplotlib.ticker import EngFormatter
+
+
+def identity(val):
+    r"""
+        returns the value as-is
+    """
+    return val
+
+
+def get_integer(val: str) -> int:
+    r"""
+        interprets the str 'val' as an integer and returns it
+    """
+    if is_integer(val):
+        return int(val)
+    else:
+        raise ValueError("Casting a non integer value: ", val)
+
+
+def is_integer(s: str) -> bool:
+    r"""
+        tests if 's' is an integer and returns a bool
+    """
+    try:
+        int(s)
+        return True
+    except ValueError:
+        return False
+
+
+def get_float(val):
+    r"""
+        interprets the str 'val' as a float and returns it
+    """
+    if is_float(val):
+        return float(val)
+    else:
+        return float("nan")
+
+
+def is_float(s: str) -> bool:
+    r"""
+        tests if 's' is an float and returns a bool
+    """
+    try:
+        float(s)
+        return True
+    except ValueError:
+        return False
+
+
+class Database:
+    r"""
+        A class that contains all the csv files within
+        the folder that it is instanced with
+    """
+    def __init__(self,
+                 data_folder_path=None,
+                 csv_separator=" ",
+                 filename_var_separator="|"):
+
+        self.datafiles = []
+        self.result_names_col = None
+        self.result_values_col = None
+        self.sim_settings_names_col = None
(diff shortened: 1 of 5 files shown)


AdelKS:
@jackun @stephanlachnit I removed the check for the python runtime dependency altogether and just wrote a comment on the `meson.build` in the `bin` folder so packagers see it.

bin/mangoplot.py
```diff
@@ -0,0 +1,376 @@
...
+    else:
+        return float("nan")
+
+
+def is_float(s):
+    try:
+        float(s)
+        return True
+    except ValueError:
+        return False
```
evelikov, line 49: Usually I don't write python, so take my comments with a grain of salt.

The above helpers should probably be `isinstance()` or wrappers around it, right?

Have you run this script through a linter - pylint, flake8, mypy to name a few - pylint throws flags 50+ issues.

bin/meson.build
```diff
@@ -0,0 +1,4 @@
+if get_option('mangoplot')
+    # runtime dependencies for the script: matplotlib and a GUI backed like PyQt5
+    install_data('mangoplot.py', install_dir: get_option('bindir'), rename: 'mangoplot', install_mode: 'rwxr-xr-x')
```
evelikov, line 3: The `install_data()` formatting is off - see the other `meson.build` for details.

The extra `option()` seems like an overkill - this is a simple python script which people can `rm` if they feel strongly against it.

AdelKS: Should be good now

bin/mangoplot.py
AdelKS: The problem these helpers are trying to solve is to know if a given `str` can be converted to `float` or `integer`. `isinstance` in this case cannot help as checks the type of the variable, in this case `s` is of type `str` :D

Yes, `flake8` gives me 5 line too long warnings that I decided to ignore, I will try pylint to see !

AdelKS: Okay I fixed some of the things that `pylint` reported (the rest is like not following naming scheme for variables). We should be good now, please feel free to comment on anything you find weird/interesting.

AdelKS:
I used a proper argument parser that gives that print a help message with `-h` and that kind of fluff, I found that to be lacking.

@NoXPhasma I took into account your feedback, please test again: further feedback welcome !

NoXPhasma:
Works fine now, dots are not ignored anymore, and the summary files are ignored.

One request I would have for future command line arguments: background and legend text color. That would allow fitting the graphs into any website design. Or, if possible, it could be changeable in the preview window.

flightlessmango:
This should have a meson option to disable/enable it. It can be enabled by default

flightlessmango:
Yeah I think you're right. There doesn't need to be an option

PR #986, opened 2023-04-13
Add missing division for fallback AMD CPU power calculation.

oliverchang:


oliverchang:
@sfrcorne @flightlessmango I noticed the latest build reporting very large CPU watt numbers on my 6800u (where previously it just failed and always returned 65W). It looks like the fallback in https://github.com/flightlessmango/MangoHud/commit/a98a40261910d417269f0c0304e2ce1a8b901c8c wasn't fully correct. 

Please let me know if this looks correct. Thanks!

flightlessmango:
Yes, I would assume this is correct

sfrcorne:
This looks indeed correct. Thanks for finding and fixing this bug in my code.

PR #1034, opened 2023-05-30
build: Add dependencies for mangoplot

Billli11:
`mangoplot` need `numpy` and `matplotlab` to run.

    import numpy as np
    
    import matplotlib.pyplot as plt
    from matplotlib.widgets import Cursor
    
    from matplotlib.ticker import EngFormatter

This add those to `build_deps.sh` and `pkgbuild/PKGBUILD`


AdelKS:
Hello, thanks for your PR ! `mangoplot` also needs a graphical front-end to `matplotlib` like `python-pyqt5`

Billli11:
@AdelKS

For `matplotlib` backends. those are dependencies of dependencies, and `mangoplot` didn't specify which one to use.

Should those needed to be added in dependencies list?



As for `ttf-lato` wouldn't it better to be included in #1033 instead?

You can find the package name with by searching each distro's package directories

[Debain](https://packages.debian.org/search?keywords=fonts-lato&searchon=names&suite=all&section=all)

[Ubuntu](https://packages.ubuntu.com/search?suite=default&section=all&arch=any&keywords=fonts-lato&searchon=names)

[Feodra](https://packages.fedoraproject.org/pkgs/lato-fonts/lato-fonts/)

[Solus](https://dev.getsol.us/source/font-lato-ttf/)

[Arch](https://archlinux.org/packages/?sort=&q=ttf-lato&maintainer=&flagged=)

AdelKS:
Thanks for your input @Billli11 

> For matplotlib backends. those are dependencies of dependencies, and mangoplot didn't specify which one to use.
>
> Should those needed to be added in dependencies list?

`matplotlib` doesn't  necessarily need a graphical backend as it can generate graphs into files directly without one, if it's only MangoHud who pulls the `python-matplotlib` package, then it definitely needs to pull a graphical backend, otherwise the script wouldn't work as-is :thinking: Is there a way to say "at least one of these dependencies must be pulled" ?

> As for ttf-lato wouldn't it better to be included in https://github.com/flightlessmango/MangoHud/pull/1033 instead?
> You can find the package name with by searching each distro's package directories [...]

I absolutely agree, I just wanted to get your help since I wasn't sure how to do with other distros. Thanks for the leads, I will try to mirror your changes in my PR then.

PR #1042, opened 2023-06-07
Gamescope: Added parameters for scaler

Root-Core:
As Gamescope has also implemented NIS, a single FSR output is really limited. They might add even more filters in the future, so it would be nice to have a singular output for all filters.

Also there is currently no output for the used scaler, which will be a easier to set option if https://github.com/ValveSoftware/gamescope/pull/893 gets merged.

I have repurposed the FSR field, please forgive me. I think a double inclusion of FSR would be wasteful.
Feedback regarding this situation is appreciated and necessary. Thank you!

### DIFF
--- a/src/app/mangoapp_proto.h
+++ b/src/app/mangoapp_proto.h
@@ -10,7 +10,7 @@ struct mangoapp_msg_v1 {
 
     uint32_t pid;
     uint64_t visible_frametime_ns;
-    uint8_t fsrUpscale;
+    uint8_t scaler_filter;
     uint8_t fsrSharpness;
     // For debugging
     uint64_t app_frametime_ns;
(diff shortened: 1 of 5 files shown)


src/app/mangoapp_proto.h
```diff
@@ -10,13 +10,14 @@ struct mangoapp_msg_v1 {
 
     uint32_t pid;
     uint64_t visible_frametime_ns;
-    uint8_t fsrUpscale;
-    uint8_t fsrSharpness;
+    uint8_t scaler_filter;
+    uint8_t scaler_sharpness;
```
flightlessmango, line 14: We don't want to rename fields even if they won't be used, add these to the end of the struct instead

flightlessmango: Yes they should still be set by gamescope with the old data for backward compatibility.
If we want to remove fields it would need to be done by creating a v2 of the struct

PR #1057, opened 2023-06-26
Update Vulkan Headers

Snowiiii:


flightlessmango:
What is the motivation to bump vulkan headers?

flightlessmango:
We don't want to bump dependencies unless they are necessary or needed for a new feature

Snowiiii:
I understand the concern about avoiding unnecessary bumps in dependencies. However, there are valid reasons to consider updating dependencies, even if they are not directly related to new features. Here are a few points to consider:

   **Bug fixes and security patches:** Newer versions of dependencies often include bug fixes and security patches. By updating dependencies, we ensure that our project remains secure and stable.

    **Performance improvements:** Dependency updates can bring performance improvements, optimizations, and better resource utilization. These improvements can enhance the overall performance of our project.

    **Compatibility with other dependencies:** Sometimes, updating a dependency is necessary to maintain compatibility with other dependencies or the underlying framework. It helps avoid conflicts and ensures a smooth integration of different components.

    **Community support:** Updating dependencies allows us to take advantage of the active community support and ongoing maintenance of the dependencies. This ensures that we have access to the latest developments and fixes provided by the community.

While it's important to exercise caution when updating dependencies, it's equally essential to strike a balance between stability and leveraging the benefits that updated dependencies offer. By evaluating the specific changes and weighing the benefits against potential risks, we can make informed decisions about when to update dependencies.

PR #1060, opened 2023-07-02
Remove debug printf in HudElements::battery

czak:
This `printf` currently keeps printing `1` or `0` depending on the `battery_watt` setting.
I belive it was just forgotten here.


czak:
Thank you!

PR #1063, opened 2023-07-07
Add ability to toggle through presets

### LINKED ISSUE #1048: Feature Request: Togglable presets
I'd like to have a feature similar to the steam deck's that allows the user to toggle between presets.  The idea is that it would work similarly to how both `toggle_fps_limit` and `fps_limit` work.  Instead of having the `preset` config option, we'd have `presets` (notice the extra s) which would be a list of togglable presets.  Additionally, like `toggle_fps_limit`, there'd be another config option `toggle_preset` that would bind the specified key to toggle presets.  Since the `preset` option is already in use, we could use that as the first and default option regardless of what's in `presets`.  Thoughts?

ForTheReallys:
Allow the user to toggle presets with a keybind similar to `toggle_fps_limit`.  This change updates the `preset` option to a list of presets instead of a single preset.  Just like with `fps_limit`, when loading the config for mangohud, the first listed preset is chosen, and just like with `fps_limit`, when the user presses the keybind associated with `toggle_preset`, the next (or the first if the end of the list is reached) preset is loaded.

Closes #1048

### DIFF
--- a/README.md
+++ b/README.md
@@ -387,7 +387,7 @@ Parameters that are enabled by default have to be explicitly disabled. These (cu
 | `permit_upload`                    | Allow uploading of logs to Flightlessmango.com                                        |
 | `picmip`                           | Mip-map LoD bias. Negative values will increase texture sharpness (and aliasing). Positive values will increase texture blurriness `-16`-`16` |
 | `position=`                        | Location of the HUD: `top-left` (default), `top-right`, `middle-left`, `middle-right`, `bottom-left`, `bottom-right`, `top-center`, `bottom-center` |
-| `preset=`                           | Set a pre-defined preset to be used, default is `-1`. Available presets:<br>`0` (No Hud)<br> `1` (FPS Only)<br> `2` (Horizontal)<br> `3` (Extended)<br> `4` (Detailed)<br>Own presets can be defined by creating a [presets.conf](data/presets.conf) file in `~/.config/MangoHud/`.                      |
+| `preset=`                          | Comma separated list of one or more presets. Default is `-1,0,1,2,3,4`. Available presets:<br>`0` (No Hud)<br> `1` (FPS Only)<br> `2` (Horizontal)<br> `3` (Extended)<br> `4` (Detailed)<br>User defined presets can be created by using a [presets.conf](data/presets.conf) file in `~/.config/MangoHud/`.                      |
 | `procmem`<br>`procmem_shared`, `procmem_virt`| Displays process' memory usage: resident, shared and/or virtual. `procmem` (resident) also toggles others off if disabled |
 | `ram`<br>`vram`                    | Display system RAM/VRAM usage                                                         |
 | `read_cfg`                         | Add to MANGOHUD_CONFIG as first parameter to also load config file. Otherwise only `MANGOHUD_CONFIG` parameters are used |
@@ -405,6 +405,7 @@ Parameters that are enabled by default have to be explicitly disabled. These (cu
 | `throttling_status`                | Show if GPU is throttling based on Power, current, temp or "other" (Only shows if throttling is currently happening). Currently disabled by default for Nvidia as it causes lag on 3000 series |
 | `time`<br>`time_format=%T`         | Display local time. See [std::put_time](https://en.cppreference.com/w/cpp/io/manip/put_time) for formatting help. NOTE: Sometimes apps may set `TZ` (timezone) environment variable to UTC/GMT |
 | `toggle_fps_limit`                 | Cycle between FPS limits. Defaults to `Shift_L+F1`                                    |
+| `toggle_preset`                    | Cycle between Presets. Defaults to `Shift_R+F10`                                      |
 | `toggle_hud=`<br>`toggle_logging=` | Modifiable toggle hotkeys. Default are `Shift_R+F12` and `Shift_L+F2`, respectively   |
 | `toggle_hud_position`              | Toggle MangoHud postion. Default is `R_Shift+F11`                                     |
 | `trilinear`                        | Force trilinear filtering                                                             |
--- a/src/keybinds.cpp
+++ b/src/keybinds.cpp
@@ -9,6 +9,7 @@ void check_keybinds(struct overlay_params& params, uint32_t vendorID){
    auto now = Clock::now(); /* us */
    auto elapsedF2 = now - last_f2_press;
    auto elapsedFpsLimitToggle = now - toggle_fps_limit_press;
+   auto elapsedPresetToggle = now - toggle_preset_press;
    auto elapsedF12 = now - last_f12_press;
    auto elapsedReloadCfg = now - reload_cfg_press;
    auto elapsedUpload = now - last_upload_press;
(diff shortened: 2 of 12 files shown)


src/overlay_params.cpp
```diff
@@ -160,8 +161,33 @@ parse_string_to_keysym_vec(const char *str)
 #define parse_upload_log(x)            {}
 #define parse_upload_logs(x)           {}
 #define parse_toggle_fps_limit(x)      {}
+#define parse_toggle_preset(x)         {}
 #endif
 
+// NOTE: This is NOT defined as an OVERLAY_PARAM and will be called manually
+static std::vector<int>
+parse_preset(const char *str)
```
flightlessmango, line 169: Why is this function necessary and why can't we use the normal params methods?

src/overlay_params.cpp
```diff
@@ -691,16 +719,19 @@ parse_overlay_config(struct overlay_params *params,
 
 #ifdef _WIN32
    params->toggle_hud = { VK_F12 };
+   params->toggle_preset = { VK_F5 };
```
flightlessmango, line 722: This should correspond to the x11 keybind and be F10

src/overlay_params.cpp
```diff
@@ -674,14 +698,18 @@ static void set_param_defaults(struct overlay_params *params){
 
 void
 parse_overlay_config(struct overlay_params *params,
-                  const char *env)
+                  const char *env, bool use_existing_preset)
 {
-   *params = {};
+   std::vector<int> default_preset = {-1, 0, 1, 2, 3, 4};
+   *params = {
+     .preset = use_existing_preset ? params->preset : default_preset
```
flightlessmango, line 705: Won't this be overwritten if preset exists in the config?

src/overlay_params.cpp
```diff
@@ -674,14 +698,18 @@ static void set_param_defaults(struct overlay_params *params){
 
 void
 parse_overlay_config(struct overlay_params *params,
-                  const char *env)
+                  const char *env, bool use_existing_preset)
 {
-   *params = {};
+   std::vector<int> default_preset = {-1, 0, 1, 2, 3, 4};
```
flightlessmango, line 703: this doesn't account for user defined presets. There's no hard limit to how many can be defined

src/keybinds.cpp
```diff
@@ -50,6 +51,19 @@ void check_keybinds(struct overlay_params& params, uint32_t vendorID){
       }
    }
 
+   if (elapsedPresetToggle >= keyPressDelay &&
+       keys_are_pressed(params.toggle_preset)) {
+     toggle_preset_press = now;
+     size_t size = params.preset.size();
```
flightlessmango, line 57: this seems over-engineered. Consider this instead:
Add default to the presets switch
```
      default:
         params->preset = 0;
         presets(0, params);
```
This way way we don't need extra variables like `current_preset` and we don't need to check if preset is valid.
Just use `params->preset` and `++` it. If that makes sense?

src/overlay_params.cpp
ForTheReallys: Yes that's the idea

src/overlay_params.cpp
ForTheReallys: Correct.  These just correspond with the current default presets found in `presets()`

src/overlay_params.cpp
ForTheReallys: I didn't use the normal params methods because the a parameters specified in a preset should be set before any other parameters are set.  This is how presets work today

src/keybinds.cpp
ForTheReallys: That would only work if the user doesn't specify a preset variable in the config file.  If it's in the config file, `presets()` returns before ever getting to the switch.  The idea behind this change is to allow the user to specify which presets to cycle through.  For example, if I set presets to `preset=3,5,2`, 3 is the default preset, and when I hit the toggle presets keybind, it cycles to 5, then 2, and then back to 3

src/overlay_params.cpp
ForTheReallys: I noticed this wasn't the case for some of the other keybinds so I just made it F5 since the last one was F4.  I'll change it to F10 though

src/overlay_params.cpp
ForTheReallys: The user can specify how every many presets he wants

README.md
```diff
@@ -404,6 +404,7 @@ Parameters that are enabled by default have to be explicitly disabled. These (cu
 | `throttling_status`                | Show if GPU is throttling based on Power, current, temp or "other" (Only shows if throttling is currently happening). Currently disabled by default for Nvidia as it causes lag on 3000 series |
 | `time`<br>`time_format=%T`         | Display local time. See [std::put_time](https://en.cppreference.com/w/cpp/io/manip/put_time) for formatting help. NOTE: Sometimes apps may set `TZ` (timezone) environment variable to UTC/GMT |
 | `toggle_fps_limit`                 | Cycle between FPS limits. Defaults to `Shift_L+F1`                                    |
+| `toggle_fps_limit`                 | Cycle between Presets. Defaults to `Shift_R+F10`                                      |
```
flightlessmango, line 407: should be `toggle_preset`?

ForTheReallys: Yes it should

flightlessmango:
Can you address the github action compile errors?

ForTheReallys:
Should be good now

ForTheReallys:
@flightlessmango bump

flightlessmango:
Can you squash the "fixes" into [Add ability to toggle through presets](https://github.com/flightlessmango/MangoHud/pull/1063/commits/c5aa6d4f2a3682fdb2630afefc54a64c209fae71) and [Update README.md](https://github.com/flightlessmango/MangoHud/pull/1063/commits/7d2a9a946ea42e0021f9c2eb1d1e729e89074a1c)?

src/keybinds.cpp
flightlessmango: how about
```current_preset = params.preset[++current_preset % size];```?

ForTheReallys:
I squashed and pushed my changes but the build failed in the configure step.  `ERROR: Program 'glslangValidator' not found or not executable` Any ideas @flightlessmango?

evelikov:
You want to rebase against master - the error has already been fixed with https://github.com/flightlessmango/MangoHud/commit/e60b625fd025518128f3cfae08e815a7f9d9642c

evelikov:
It seems to me that "Set keybinding for toggling presets to F10 on windows" should be squashed with "Add ability to toggle through presets"

ForTheReallys:
Bump @flightlessmango

flightlessmango:
I'm not happy with how it needs to bend the param systems to work.
This is a design flaw in how I wrote presets.
I'm not sure how I want to proceed yet

ForTheReallys:
Any updates on this @flightlessmango?

flightlessmango:
I noticed after merge that if you don't set a preset, it defaults to preset 0.
So the hud is hidden by default

flightlessmango:
Another problem is that if you set preset and then unset preset, it will remember the last preset.
This shouldn't happen

flightlessmango:
I believe I've resolved the issues in 02c3b4f

flightlessmango:
spoke too soon, that breaks toggling presets

ForTheReallys:
> I noticed after merge that if you don't set a preset, it defaults to preset 0. So the hud is hidden by default

Fixing and creating another PR shortly

> Another problem is that if you set preset and then unset preset, it will remember the last preset. This shouldn't happen

Can you elaborate on this?

ForTheReallys:
> I noticed after merge that if you don't set a preset, it defaults to preset 0. So the hud is hidden by default

> spoke too soon, that breaks toggling presets

Both are fixed in #1115

PR #1071, opened 2023-07-18
Remove nlohmann_json from DFSG tarball

stephanlachnit:


flightlessmango:
How does this go against DFSG?

flightlessmango:
Oh I'm just being stupid, nevermind!

stephanlachnit:
> How does this go against DFSG?

It's not, we just that we are trying to avoid embedded source copies where possible.

Issue #1062, opened 2023-07-05
Feature: amdgpu support for temp1-3 [PATCH]


axet:
**Describe the bug**

amd video card has multiply temp sensors. By default mangohud only using temp1_input which is not optional. Since this is not a maximum temperature. According to 'lm_sensors' package maximum temperature is 'junction' which is corresponding to 'temp2_input' sensor.

On my video card (RX 6600) temp1_input around 60 when temp2_input about 80.

I suggest add MAX from all 3 temp sensors to show user maxmium temperature on video card, not 1 sensor.

**List relevant hardware/software information**

- debian bookworm
- MangoHud 0.6.8-2
- GPU RX 6600

**Expected behavior**

mangohud show maximum temp. not first sensor values.

**Additional context**

* https://gitlab.com/axet/homebin/-/blob/debian/cdebian-dget.d/bookworm/mangohud/amdgpu_3_temp.patch

flightlessmango:
I'm thinking this and removing `gpu_temp_junction` as well. Could you PR this?

siroccal:
> I'm thinking this and removing `gpu_temp_junction` as well. Could you PR this?

No, you should keep the normal/average die temperature and junction temperature separate. They're not the same thing and they should not be treated the same. It is normal for the junction temperature to go higher than die temperature: https://www.techpowerup.com/258228/110-c-hotspot-temps-expected-and-within-spec-amd-on-rx-5700-series-thermals.
AMD's own performance metric software also shows both temperatures separately: https://www.amd.com/en/support/kb/faq/dh3-038.

axet:
Basically you are saying - I want all three temp but I wont tell you which video card has which temperature limits on every temp sensor. I do not think MangoHud should be treated as system monitor software. Showing maximum temperature is the one way showing minimum information on screen with most important values. Showing all three values, or showing first one, without helping user to understand it's limits - uninformative.

Since I created patch for mango 0.6.8, which only showing first temp sensor. I suggested to show maximum temperature.

Recent dev branch already showing three temps sensors, we do not need to change anything. Me? I would prefer less screen pollution with one maximum temp sensor value.

siroccal:
> I wont tell you which video card has which temperature limits on every temp sensor

Every GPU has different temperature limits and the normal/average die limit is different from the junction (hotspot) temperature limit. Older GPU's often don't even have a junction temperature measurement diode (or memory temperature).

[...]

PR #1077, opened 2023-07-23
using MAX temp for amdgpu temps, drop temp separation (mem and tunction)

axet:
Closing #1062

### DIFF
--- a/src/gpu.cpp
+++ b/src/gpu.cpp
@@ -81,6 +81,7 @@ nvapi_util();
 #endif
 }
 
+#define MAX(x, y) (((x) > (y)) ? (x) : (y))
 void getAmdGpuInfo(){
 #ifdef __linux__
     int64_t value = 0;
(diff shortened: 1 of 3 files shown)


flightlessmango:
Looks good, can you just squash the commits?

axet:
ok. Gitlab has "squish commits" during the merging pull request. Strange github does't have it.

flightlessmango:
It does exist but we have it disabled for *some* reason

axet:
I also removed memory_temp and junction_temp. I guess I had to remove those from params and hud

flightlessmango:
This makes `gpu_junction_temp` and `gpu_mem_temp` non-functional.
I would say make them functional again and if `gpu_junction_temp` is enabled then `gpu_temp` is temp1 and `gpu_junction_temp` is junction like it originally worked.

siroccal:
This is a bad PR, because normal die temperature and junction temperature are both valuable information https://github.com/flightlessmango/MangoHud/issues/1062#issuecomment-1672380816. They're not the same thing and they should not be treated as if they're the same, because it's normal for junction temperature to be higher. If you want to see your junction temperature, just enable `gpu_junction_temp`...

> This makes `gpu_junction_temp` and `gpu_mem_temp` non-functional. I would say make them functional again and if `gpu_junction_temp` is enabled then `gpu_temp` is temp1 and `gpu_junction_temp` is junction like it originally worked.

No, it should always be clear what input is being displayed. If you mix data (e.g. with `max(gpu_temp, gpu_junction_temp)`), then you don't know if your average die temperature is too high even if junction temperature is not too high.

The way it works now (`gpu_temp` = `temp1_input`, `gpu_junction_temp` = `temp2_input`, and `gpu_mem_temp` = `temp3_input`) is good and should not be changed.

PR #1086, opened 2023-08-01
Show `DXVK_FRAME_RATE` limit on hud if set

HeroBrine1st:
Check if DXVK_FRAME_RATE is set and uses it to show instead of fps limit:

image

This change is purely visual. Of course it does not check DXVK config.

Tested on Elite: Dangerous in following combinations:

1. No `DXVK_FRAME_RATE` variable set and no config limit - shows `Fps limit [late]: 0`
2. `DXVK_FRAME_RATE=75` and no config limit - shows like in screenshot
3. With `fps_limit=60` in config and `DXVK_FRAME_RATE=75` - `Fps limit [late]: 60`
4. The same 3 but with `fps_limit_method=early` - works as expected
5. The same 3 but with `fps_limit_method=late` - works as 3

You may think this implementation is rather dirty and environment read should be decoupled from the hud element. If so, you can simply close this PR and think it is a feature request, because C++ is not my main language and thus I'm not able to deep dive in this project in reasonable amount of time.


flightlessmango:
This should check if we're using dxvk by checking `HUDElements.sw_stats->engine == EngineTypes::DXVK`.
If both mangohud and dxvk has fps limit it should display which ever has the lowest limit set.
Not sure how to handle if both have the same limit, probably show dxvk?

HeroBrine1st:
I think in the same limit case we should show dxvk because in late case mangohud presents frame and waits with dxvk, then mangohud requests frame, dxvk creates it and mangohud instantly presents it, or in early case mangohud, I guess, gets frame, instatly shows it and waits for next frame from dxvk. But I may be wrong, I read about late/early limiter like 30 minutes ago.

I implemented dxvk engine check and ^, btw.

HeroBrine1st:
Added a fix so that it still shows "Fps limit [late]" if DXVK_FRAME_RATE is set to invalid value. Sorry for interrupting actions.

flightlessmango:
Can you squash the commits? Otherwise I think we're good

HeroBrine1st:
Also I found a bug by "squash verification" review: if there's no fps_limit in config, condition for showing dxvk won't be true (I tested it before fixing of course), so `if(fpsDxvk > 0 && fpsDxvk <= fps)` -> `if(fpsDxvk > 0 && (!fps || fpsDxvk <= fps))`. Squashed it as well and tested again: no fps_limit in config and it shows dxvk limit now.

Issue #1038, opened 2023-06-02
Feature: single / total cores cpu load


axet:
**Describe the bug**

Some games has single core engines and it loads 100% of single core. Bug total cpu usage become 20% which is not accurate since game can only utilize one core.

**List relevant hardware/software information**

- debian bookworm
- MangoHud 0.6.8-2

**Expected behavior**

allow to show single core cpu load + total load like this:

CPU 100% 20% 45°

axet:
Right we can not know if game engine only using one core. But from user perspective showing single core usage and total usage would be more accurate.

here is a patch

https://gitlab.com/axet/homebin/-/blob/debian/cdebian-dget.d/bookworm/mangohud/cpusingle.patch

axet:
For example game showing 30fps with 60fps limit. CPU usage 20% GPU usage  20%

From user perspective this makes no sense. What causing low FPS? We have to show a single core load (maximum load from all cores)

flightlessmango:
You're making an assumption that the highest load on a core is from the game, which again we can't know. It could be anything. If the user happens to be running something in the background it will be even more confusing

coffeecrisp:
I suppose it can be done since System Monitor in KDE can show a list of all processes and % CPU usage.
Then, you just do some math.
If the process uses 20% and you have 16 cores:
20% of 16 = 3.2 cores <==== we approximate to 3 cores.

MangoHud would display ===> This game is using 3 cores.

I am not an expert in Linux, so I do not know how to get that percentage for a process.

axet:
We can call it compact version of 'core_load=1'. Modern CPU has a lot of cores (8-12). Showing all of them takes a lot of space on screen, and not very informative. Simply compact all cores into one maximum load core + showing total load can achieve the same result with one/two lines of text.

coffeecrisp:
You want MangoHud to show only cores that are at 100% usage?
Can you write it as an algorithm?

axet:
Algorithm? patch above

flightlessmango:
I don't think this is a good idea

JFingerle:
I know that this issue has been closed, but this would be a really good feature to detect a CPU bottleneck.

axet:
Use patch from above. Histogram pull request does not cover it.

JFingerle:
> Use patch from above. Histogram pull request does not cover it.

Thank you, but the patch in your gitlab repo is missing :(

axet:
[...]

PR #1091, opened 2023-08-11
Add core_bars for a per CPU core load histogram

jansol:
Decided to take a stab at the core histogram I mentioned in #1038 

Not sure if it solves OP's problem there, but for me I find that it helps get a rough idea of the structure of the CPU load.

Screenshot

### DIFF
--- a/README.md
+++ b/README.md
@@ -317,6 +317,7 @@ Parameters that are enabled by default have to be explicitly disabled. These (cu
 | `control=`                         | Sets up a unix socket with a specific name that can be connected to with mangohud-control.<br>I.e. `control=mangohud` or `control=mangohud-%p` (`%p` will be replaced by process id)    |
 | `core_load_change`                 | Change the colors of cpu core loads, uses the same data from `cpu_load_value` and `cpu_load_change` |
 | `core_load`                        | Display load & frequency per core                                                     |
+| `core_bars`                        | Change the display of `core_load` from numbers to vertical bars                       |
 | `cpu_load_change`                  | Change the color of the CPU load depending on load                                    |
 | `cpu_load_color`                   | Set the colors for the gpu load change low, medium and high. e.g `cpu_load_color=0000FF,00FFFF,FF00FF` |
 | `cpu_load_value`                   | Set the values for medium and high load e.g `cpu_load_value=50,90`                    |
--- a/src/hud_elements.cpp
+++ b/src/hud_elements.cpp
@@ -348,10 +348,50 @@ void HudElements::cpu_stats(){
     }
 }
 
+
+static float get_core_load_stat(void*,int);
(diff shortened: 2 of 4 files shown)


flightlessmango:
Look goods. The only thing I'm thinking about is we want this to be displayed with `full` or not. Seems like overkill to have `core_load` and `core_bars` in `full`

jansol:
Yeah having both is definitely unnecessary and which one should be there by default is up for debate. On one hand the numbers are more precise but the bars are better at giving an overview. I'm fine with any arbitrary decision in that regard - `full` is ultimately just a convenience macro. I guess the main question is how stable people expect `full` to be across versions?

Another option that would be cool would be to draw the load numbers right on top of the bars (vertically?) and use that to replace `core_load` completely, but I am not sure if imgui lets us do that without an unnecessary amount of work(arounds).

flightlessmango:
Yeah let's not change the `full` for now. 

Can you set core_bars to false here          https://github.com/flightlessmango/MangoHud/blob/41f923b0b2be07f035df221d01625e11f1305fde/src/overlay_params.cpp#L733

and here
https://github.com/flightlessmango/MangoHud/blob/41f923b0b2be07f035df221d01625e11f1305fde/src/overlay_params.cpp#L544

jansol:
I've been thinking, maybe it makes most sense to have this as a modifier option that is off by default like `histogram` for frametime?

flightlessmango:
You need to set core_bars to false here

https://github.com/flightlessmango/MangoHud/blob/08152325d4b8a168dab460c0afd1d49e29cea2d8/src/overlay_params.cpp#L568

and also here
https://github.com/flightlessmango/MangoHud/blob/08152325d4b8a168dab460c0afd1d49e29cea2d8/src/overlay_params.cpp#L752

Otherwise it will be enabled with `full`

Also it appears core_bars is enabled by default? As just setting `core_load` gives the bars

jansol:
Should work now. At least core_load alone gives me numbers, core_bars alone does nothing and adding both gives me the bars as expected.

jansol:
I ran into a segfault somewhere in dbus when trying with `full` so couldn't verify that one but somehow I don't think that is related to these changes.

axet:
Should be visually different 100% load from 99% load per core. By color, number, shape...

jansol:
> Should be visually different 100% load from 99% load per core. By color, number, shape...

Yeah I would have loved that as well but didn't find an easy way to do it in the imgui histogram function. If someone finds a way or wants to take the time to write a custom drawing routine for this I would love to see it. To me personally it's not important enough to justify spending a lot of time on it, though. The current implementation already tells me what I want to know.

flightlessmango:
> By color, number, shape

This is probably doable using ImPlot which is included in mangohud

PR #1093, opened 2023-08-17
Implement wayland hotkeys

Etaash-mathamsetty:
ready for review!

### DIFF
--- a/.github/workflows/ubuntu.yml
+++ b/.github/workflows/ubuntu.yml
@@ -28,7 +28,8 @@ jobs:
           libglew-dev \
           libglfw3-dev \
           libwayland-dev \
-          libxnvctrl-dev
+          libxnvctrl-dev \
+          libxkbcommon-dev
         sudo pip3 install 'meson>=0.60'
     - name: 'Install clang'
       if:  ${{ (matrix.compiler == 'clang') }}
--- a/src/keybinds.cpp
+++ b/src/keybinds.cpp
@@ -1,9 +1,15 @@
 #include <cstdint>
+#include <cstring>
+#include <array>
+#include <algorithm>
+#include <unistd.h>
 #include "overlay.h"
 #include "timing.hpp"
 #include "logging.h"
 #include "keybinds.h"
 #include "fps_metrics.h"
+
+Clock::time_point last_f2_press, toggle_fps_limit_press, toggle_preset_press, last_f12_press, reload_cfg_press, last_upload_press;
 
 void check_keybinds(struct overlay_params& params, uint32_t vendorID){
    using namespace std::chrono_literals;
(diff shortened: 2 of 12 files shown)


src/keybinds.cpp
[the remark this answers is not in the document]
Etaash-mathamsetty: I agree we need to split it, but the reason wayland_hook is separate is because it's literally not possible for them to be in the same file due to redefinitions

src/wayland_hook.h
```diff
@@ -0,0 +1,2 @@
+extern void* wl_handle;
+extern void* wl_display_ptr;
```
i509VCB, line 2: There are clients which connect to the same display multiple times. Does this need to consider that?

orowith2os: fwiw it might be good to look at how a similar situation is handled on X11, if possible?

Etaash-mathamsetty:
hooking up vulkan will require us to use elf and replace every instance of wl_display_flush, im not interesting in doing that :/

jansol:
Maybe have a look at what renderdoc's capture layer does? It also registers a hotkey that bypasses the app where the layer is loaded.

i509VCB:
> Maybe have a look at what renderdoc's capture layer does? It also registers a hotkey that bypasses the app where the layer is loaded.

This would require intercepting messages. Maybe something using libei could work? Although that would still require compositor support

Etaash-mathamsetty:
> libei

does plasma 5.27 support libei?

Etaash-mathamsetty:
> hooking up vulkan will require us to use elf and replace every instance of wl_display_flush, im not interesting in doing that :/ there's got to be a way of setting LD_PRELOAD in a vulkan layer

im a stupid guy in a world that has already done this before :)

Etaash-mathamsetty:
missing things:
1. support for multiple vulkan displays
2. ifdefs for wayland
3. shifting some of the wayland keyboard stuff out of keybinds.cpp

src/wayland_hook.h
Etaash-mathamsetty: > There are clients which connect to the same display multiple times. Does this need to consider that?

not sure but it seems to work with everything that I have tried so far

flightlessmango:
Can you squash the unnecessary commits?

flightlessmango:
> yea should I squash all of them into one or?

Yeah I think that would be fine

.github/workflows/ubuntu.yml
```diff
@@ -28,7 +28,8 @@ jobs:
           libglew-dev \
           libglfw3-dev \
           libwayland-dev \
-          libxnvctrl-dev
+          libxnvctrl-dev \
+          libxkbcommon-dev
```
flightlessmango, line 32: This dep should be added to [build_deps.sh](https://github.com/flightlessmango/MangoHud/blob/master/build_deps.sh)

arch - `libxkbcommon`
fedora - `libxkbcommon-devel`
solus - `libxkbcommon-devel`
suse - `libxkbcommon-devel`

PR #1115, opened 2023-09-09
Default to -1 preset instead of 0 when preset is unset

ForTheReallys:
Default to -1 preset if `preset` is not set and fix toggling issue caused by [02c3b4f](https://github.com/flightlessmango/MangoHud/commit/02c3b4f235a2ff5581220b2cd1aa34a50dd20bcf)


flightlessmango:
Looks like everything works as expected now, thanks for fixing this so quickly :+1:

PR #1121, opened 2023-09-12
Parsing fixes for presets.conf

hannesmann:
1. Use `std::getline` instead of `fgets`. Long lines could overflow the buffer which only held 20 characters.
Fixes cases like these:
```
[preset 1]
fps_limit=140,120,60,0
```

2. Only break if we get a line that matches the pattern "[...]".
Fixes cases like these:
```
[preset 1]
fps_limit=140,120,60,0
fps_limit_method=early

cpu_stats
cpu_temp
cpu_power
cpu_mhz

gpu_stats
gpu_temp
gpu_core_clock
gpu_mem_clock
gpu_mem_temp
gpu_power
```

### DIFF
--- a/src/overlay_params.cpp
+++ b/src/overlay_params.cpp
@@ -14,6 +14,7 @@
 #include <iostream>
 #include <string>
 #include <sstream>
+#include <fstream>
 #include <algorithm>
 #include <cctype>
 #include <array>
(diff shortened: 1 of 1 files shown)


flightlessmango:
Looks good, can you squash the commits?

hannesmann:
> Looks good, can you squash the commits?

Done.

PR #1124, opened 2023-09-19
logging: include swap and RSS sizes

craftyguy:
These stats are useful for comparing memory usage over time, and any impacts to performance that might be from changing swap usage during runtime.

### DIFF
--- a/src/overlay.cpp
+++ b/src/overlay.cpp
@@ -159,6 +159,7 @@ void update_hw_info(const struct overlay_params& params, uint32_t vendorID)
 #ifdef __linux__
    currentLogData.ram_used = memused;
    currentLogData.swap_used = swapused;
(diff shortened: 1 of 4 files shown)


src/overlay.cpp
```diff
@@ -158,6 +158,8 @@ void update_hw_info(const struct overlay_params& params, uint32_t vendorID)
    currentLogData.gpu_power = gpu_info.powerUsage;
 #ifdef __linux__
    currentLogData.ram_used = memused;
+   currentLogData.swap_used= swapused;
+   currentLogData.process_rss= proc_mem.resident / float((2 << 29)); // GiB, consistent w/ other mem stats
```
flightlessmango, line 162: missing space before `=`

PR #1133, opened 2023-09-28
Fix opengl runtime on musl libc systems

Johnnynator:
- elfhacks: d_un.d_ptr is relative on non glibc systems
- meson: allow setting absolute path instead of 'Dynamic string tokens'


flightlessmango:
Can you resolve the conflict?

Johnnynator:
Done, rebased on top of master.

PR #1136, opened 2023-09-29
imgui: Disable webgpu option in meson.build

Billli11:
Since imgui have been updated to 1.89.9, new option 'webgpu' have been added to imgui's meson.build.

Keeping it set to auto can have building problem if --auto_features is enabled for meson.(like arch-meson)


flightlessmango:
Already added it here 9a0ecee

PR #1151, opened 2023-10-11
pkgbuild: update source for sdplog to 1.12.0

Billli11:


flightlessmango:
This version needs `external_fmt` set to disabled, otherwise it crashes cs2

Billli11:
> This version needs `external_fmt` set to disabled, otherwise it crashes cs2

Wouldn't it be better if it is set in `meson.build`?

Also look like external_fmt is already disable in `meson.build`

    if not spdlog_dep.found()
      spdlog_sp = subproject('spdlog', default_options: [
        'default_library=static',
        'compile_library=true',
        'werror=false',
        'tests=false',
        'external_fmt=disabled'
      ])

flightlessmango:
We're good then!

Issue #1129, opened 2023-09-27
Can't set preset via MANGOHUD_CONFIG


jarettmillard:
Setting `preset` in `~/.config/MangoHud/MangoHud.conf` works, but setting it via `MANGOHUD_CONFIG` does not.

**List relevant hardware/software information**
- Linux Distribution: Pop OS
- MangoHud version: 0.7.0

**To Reproduce**
```
$ MANGOHUD_CONFIG='preset=4' vkcube
[2023-09-27 02:46:18.131] [MANGOHUD] [error] [overlay_params.cpp:614] Unknown option 'preset'
```

**Expected behavior**
The preset should be applied.

PR #1167, opened 2023-11-15
Presets: Add ability to specify preset_path via env variables

### LINKED ISSUE #1165: [Feature Request] Allow specifying presets.conf file path in config.
**Describe the bug**
Right now it is not possible to configure presets inside Flatpak. The reason for that is MangoHud looking for hardcoded path `~/.config/MangoHud/presets.conf` which is not achievable in Flatpaks as homedir-relative paths don't work because of sandbox reasons (`~/` is not the same as `~/` outside of the Flatpak). I can specify config path with env variable like this `MANGOHUD_CONFIGFILE=/var/home/USERNAME/.config/MangoHud/MangoHud.conf` and it works, but there's no way to specify presets file path in env vars or in the config itself.

- Fedora Silverblue 39
- org.freedesktop.Platform.VulkanLayer.MangoHud//22.08 (Flatpak)

**Expected behavior**
Specifying presets.conf file path with something like `MANGOHUD_PRESETSFILE=/path/to/presets.conf` as env variable or `presets_file=/path/to/presets.conf` in MangoHud config.

ForTheReallys:
Fixes  #1165

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ A Vulkan and OpenGL overlay for monitoring FPS, temperatures, CPU/GPU load and m
   - [Normal usage](#normal-usage)
   - [OpenGL](#opengl)
   - [Hud configuration](#hud-configuration)
-    - [Environment Variables: **`MANGOHUD_CONFIG`** and **`MANGOHUD_CONFIGFILE`**](#environment-variables-mangohud_config-and-mangohud_configfile)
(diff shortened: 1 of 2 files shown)


flightlessmango:
@ForTheReallys thanks for this PR.
It would be really awesome if you could also look at fixing presets when set in `MANGOHUD_CONFIG=` #1129

ForTheReallys:
@flightlessmango already started :)

PR #1172, opened 2023-11-23
Adding command for arch in README.md

chowned:
Guide was missing the actual command to install the software in arch linux and based distros, it stopped at refreshing databases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,10 +164,16 @@ and uncomment:
 #Include = /etc/pacman.d/mirrorlist
 ```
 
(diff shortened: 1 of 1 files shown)


flightlessmango:
This should also mention 32bit as we're instructing the user to add multilib repo

chowned:
Sure thing! Should be done and should be quite friendly to follow now. Tell me if i need to change or implement something more!

chowned:
I will create a new pull request if it is not a problem

PR #1176, opened 2023-12-01
amdgpu, overlay: Use gpu_metrics for CoreClock on Steam Deck

Joshua-Ashton:
sysfs path gives 200 or 1040 Mhz always. Don't use that!

### DIFF
--- a/src/amdgpu.cpp
+++ b/src/amdgpu.cpp
@@ -250,7 +250,7 @@ void amdgpu_metrics_polling_thread() {
 	}
 }
 
-void amdgpu_get_metrics(){
+void amdgpu_get_metrics(uint32_t deviceID){
 	static bool init = false;
 	if (!init){
 		std::thread(amdgpu_metrics_polling_thread).detach();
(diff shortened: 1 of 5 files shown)


src/amdgpu.cpp
```diff
@@ -264,7 +264,12 @@ void amdgpu_get_metrics(){
 	gpu_info.MemClock = amdgpu_common_metrics.current_uclk_mhz;
 
 	// Use hwmon instead, see gpu.cpp
-	// gpu_info.CoreClock = amdgpu_common_metrics.current_gfxclk_mhz;
+	if ( deviceID == 0x1435 || deviceID == 0x163f )
+	{
+		// If we are on VANGOGH (Steam Deck), then
+		// always use use core clock from GPU metrics.
+		gpu_info.CoreClock = amdgpu_common_metrics.current_gfxclk_mhz;
```
flightlessmango, line 271: This will be overwritten in [gpu.cpp#138](https://github.com/flightlessmango/MangoHud/blob/f168483c3ef059abf6a5072ffbdd2c7c3104e39b/src/gpu.cpp#L141C11-L141C11)
Should also have a condition based on deviceID there

Joshua-Ashton: It is based on deviceID already?

It will not be overwrittenm this comes after.

flightlessmango: Ah yes you are correct, looks good then!

flightlessmango: Ah yes you are correct, looks good then!

K900:
This broke the tests :(

flightlessmango:
> This broke the tests :(

fixed here 4646e2e4f6fe3067740b9eb806679b7576231f10

PR #1177, opened 2023-12-01
overlay_params: Don't show version/time for preset 4

lostgoat:
We should probably re-add the time with some better formatting.


flightlessmango:
This has already been addressed

PR #1181, opened 2023-12-06
Correction to Arch Linux guide

chowned:
Guide was missing the actual command to install the software in arch linux and based distros, it stopped at refreshing databases.
I added 32 bit and 64 bit installer for Arch Linux and squeezed to just one commit as previously requested.

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,17 +157,28 @@ If you are building it by yourself, you need to enable multilib repository, by e
 sudo nano /etc/pacman.conf
 ```
 
-and uncomment:
+and uncomment the lines in order to look like this:
 
-```txt
-#[multilib]
-#Include = /etc/pacman.d/mirrorlist
+```ini
+[multilib]
+Include = /etc/pacman.d/mirrorlist
 ```
 
-then save the file and execute:
+then save the file and execute the command to refresh databases:
 
 ```
 sudo pacman -Syy
+```
+
+and install the 64 bit package with:
+
+```
+sudo pacman -S mangohud
+```
+
+For 32 bit package, install with:
+```
+sudo pacman -S lib32-mangohud
 ```
 
 ### Debian, Ubuntu


Doomer-Cat:
just adding mangohud + lib32-mangohud to db refresh and describing would have been sufficient in my opinion

chowned:
I followed a previous pull request from here

https://github.com/flightlessmango/MangoHud/pull/1172

Doomer-Cat:
in my opinion even this would be enough:

[Enable multilib in /etc/pacman.conf with **sudo**](https://wiki.archlinux.org/title/Official_repositories#Enabling_multilib)
Now refresh databases and install the 32-bit and 64-bit mangohud packages:
```
sudo pacman -Syy lib32-mangohud mangohud
```

flightlessmango:
> [Enable multilib in /etc/pacman.conf with **_sudo_**](https://wiki.archlinux.org/title/Official_repositories#Enabling_multilib) Now refresh databases and install the 32-bit and 64-bit mangohud packages:
> 
> ```
> sudo pacman -Syy lib32-mangohud mangohud
> ```
Something like this is probably better. We get a lot of issues that stems from the user not having 32bit installed, it would be good if users are instructed to install both at the same time

hashworks:
> 
> ```
> sudo pacman -Syy lib32-mangohud mangohud
> ```

Please don't force refresh your local package databases (`yy`) unless they are corrupt. One `y` is enough.

flightlessmango:
> I'm not totally agreeing on installing both packages but the user could just install one

I agree with you, however I'm thinking about `pacman -Syy`. I believe it's bad practice to update the database without updating the system

hashworks:
> > I'm not totally agreeing on installing both packages but the user could just install one
> 
> I agree with you, however I'm thinking about `pacman -Syy`. I believe it's bad practice to update the database without updating the system

As I said above, running `yy` is discouraged anyway - it forces a redownload of all package databases and puts unneeded strain on Arch Linux infrastructure. It should only be used to fix broken systems. A single `y` will check if there are updates and downloads only what's needed.

And yes, in general partial upgrades are unsupported and one should use `pacman -S` or `pacman -Syu` unless they know what they are doing.

PR #1209, opened 2024-01-12
Add hardware monitoring of asus boards

e2dk4r:
```shell
$ cat /sys/class/dmi/id/board_name 
ROG CROSSHAIR VIII DARK HERO
```

My kernel config
```shell
$ cat /boot/config-6.7.0
...
CONFIG_SENSORS_ASUS_EC=y
...
CONFIG_SENSORS_NCT6775_CORE=y
CONFIG_SENSORS_NCT6775=y
...

### DIFF
--- a/src/cpu.cpp
+++ b/src/cpu.cpp
@@ -519,8 +519,11 @@ bool CPUStats::GetCpuFile() {
         } else if (name == "it8603") {
             find_input(path, "temp", input, "temp1");
             break;
-        } else if (name == "nct6797") {
+        } else if (starts_with(name, "nct")) {
             find_input(path, "temp", input, "TSI0_TEMP");
+            break;
+        } else if (name == "asusec") {
+            find_input(path, "temp", input, "CPU");
             break;
         } else {
             path.clear();

### CODE REVIEW

**Comment by flightlessmango** [comment]
It seems annoying to have to add `nct6798` + 1 every iteration of the module.
Could this be simplified to just checking for `nct` instead?

**Comment by e2dk4r** [comment]
nct6775 driver supports below devices, if I am not mistaken
```c
static const char * const [nct6775_device_names](https://elixir.bootlin.com/linux/v6.7/C/ident/nct6775_device_names)[] = {
	"nct6106",
	"nct6116",
	"nct6775",
	"nct6776",
	"nct6779",
	"nct6791",
	"nct6792",
	"nct6793",
	"nct6795",
	"nct6796",
	"nct6797",
	"nct6798",
	"nct6799",
};
```
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/drivers/hwmon/nct6775-core.c?h=v6.7#n64

I would prefer to use kernel's supported names to be exact.
What you would recommend?

PR #1247, opened 2024-02-28
Add liquid cooling stats

Makevelli:
Add the possibility to display sensors data from watercooling devices. It is mainly for Aquacomputer devices since mainline kernel support is available. Support for devices from other manufacturers could be added if there are kernel modules that expose them in sysfs.

### DIFF
--- a/src/hud_elements.cpp
+++ b/src/hud_elements.cpp
@@ -20,6 +20,7 @@
 #include <IconsForkAwesome.h>
 #include "version.h"
 #include "blacklist.h"
+#include "liquid.h"
 #ifdef __linux__
 #include "implot.h"
 #endif
@@ -99,6 +100,7 @@ void HudElements::convert_colors(const struct overlay_params& params)
     HUDElements.colors.fps_value_high = convert(params.fps_color[2]);
     HUDElements.colors.text_outline = convert(params.text_outline_color);
     HUDElements.colors.network = convert(params.network_color);
+    HUDElements.colors.liquid = convert(params.liquid_color);
 
     ImGuiStyle& style = ImGui::GetStyle();
     style.Colors[ImGuiCol_PlotLines] = convert(params.frametime_color);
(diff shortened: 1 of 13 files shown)


src/liquid.h
```diff
@@ -0,0 +1,47 @@
...
+    float GetTemp();
+    std::string& GetDeviceName();
+    int GetDeviceCount() const;
+    std::vector<WatercoolingDevice> GetDevicesData() const;
+
+private:
+    std::vector<WatercoolingDevice> devices;
+};
+
+extern LiquidStats liquidStats;
```
flightlessmango, line 45: this should be a unique_ptr

src/overlay_params.cpp
```diff
@@ -721,6 +727,8 @@ static void set_param_defaults(struct overlay_params *params){
    params->table_columns = 3;
    params->text_outline_color = 0x000000;
    params->text_outline_thickness = 1.5;
+   params->liquid_color=0x3fcbd4;
```
flightlessmango, line 730: missing spaces before and after `=`

src/hud_elements.cpp
```diff
@@ -1196,6 +1203,78 @@ void HudElements::device_battery()
 #endif
 }
 
+void HudElements::liquid_stats()
```
flightlessmango, line 1206: Since liquid is only for unix and we only compile liquid.cpp on unix, the windows compiler does not have access to liquidStats.
set `#ifndef _WIN32` inside this function so it's not being compiled with mingw.
This should resolve the failing mingw test

Makevelli:
Updated to fix mingw build and apply your suggestions

GR3ee3N:
What is the status on this? Would be really cool to have my Aquacomputer sensors available 🙂

Makevelli:
Screenshot_20260326_152327
Works great on my private MangoHud version, haha. 
On a more serious note, I submitted some changes for review, however @flightlessmango no longer answered...so I kept updating it in private for my use. I don't know if it's still a wanted feature, given how few people use watercooling and even less using Aquacomputer devices.

flightlessmango:
Thanks for the PR and for the effort here.

I’m going to decline this for now because of the rewrite work in progress. The new config system should be able to handle this kind of reporting without hard-coded device-specific support, and in a way that can cover a broader range of hardware.

Because of that, I don’t think it makes sense to merge this implementation into the current codebase.

PR #1252, opened 2024-03-04
Add support for 'preset' option in MANGOHUD_CONFIG

### LINKED ISSUE #1129: Can't set preset via MANGOHUD_CONFIG

ForTheReallys:
Sorry for the long wait. Took some time to focus on some life stuff.

This PR adds support for passing 'preset' using the environment variable. It additionally has a few (hopefully desirable) side effects:

- Consistently handle options when passed through MANGOHUD_CONFIG and through the config file (e.g. "full")
- Better handle read_cfg (no longer required as the first argument in MANGOHUD_CONFIG)
-  Because of the previous inconsistency when using "full" through the env and the config file, I chose to use the same options that were previously used in the config file instead of the environment variable.

Let me know your thoughts

Fixes #1129

### DIFF
--- a/src/overlay_params.cpp
+++ b/src/overlay_params.cpp
@@ -572,60 +572,114 @@ const char *overlay_param_names[] = {
 };
 
 static void
-parse_overlay_env(struct overlay_params *params,
-                  const char *env)
+initialize_preset(struct overlay_params *params)
 {
+   if (params->options.find("preset") != params->options.end()) {
+      auto presets = parse_preset(params->options.find("preset")->second.c_str());
+      if (!presets.empty())
+         params->preset = presets;
+   }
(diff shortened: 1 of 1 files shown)


flightlessmango:
All of it seems like improvements to me. Thanks for fixing this!

PR #1262, opened 2024-03-13
Add various additional stats to log summary

alkazar:
Added various stats to the log summary, preserving the order and headers of the original CSV format for backwards compatibility.

The following stats were added:
 - Average Frame Time
 - Average GPU Temp
 - Average CPU Temp
 - Average VRAM Used
 - Average RAM Used
 - Average Swap Used
 - Peak GPU Load
 - Peak CPU Load
 - Peak GPU Temp
 - Peak CPU Temp
 - Peak VRAM Used
 - Peak RAM Used
 - Peak Swap Used

Testing performed:
 - Logged several gaming sessions
 - Imported the summary CSV file into Google Sheets
 - Validated that the format was as expected and all values seemed accurate

### DIFF
--- a/src/logging.cpp
+++ b/src/logging.cpp
@@ -1,6 +1,7 @@
 #include <sstream>
 #include <iomanip>
 #include <array>
+#include <algorithm>
 #include <spdlog/spdlog.h>
 #include "logging.h"
 #include "overlay.h"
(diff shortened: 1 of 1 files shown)


src/logging.cpp
```diff
@@ -71,12 +72,24 @@ static void writeSummary(string filename){
...
     float total = 0.0f;
-    float total_cpu = 0.0f;
     float total_gpu = 0.0f;
+    float total_cpu = 0.0f;
+    int total_gpu_temp = 0.0f;
+    int total_cpu_temp = 0.0f;
+    float total_vram = 0.0f;
+    float total_ram = 0.0f;
+    float total_swap = 0.0f;
+    int peak_gpu = 0.0f;
```
alkazar, line 86: I am not really sure why, but the compiler yelled at me when I made this a `float` type.

src/logging.cpp
```diff
@@ -91,21 +104,66 @@ static void writeSummary(string filename){
     // 97th percentile
     result = sorted.empty() ? 0.0f : 1000 / sorted[floor(0.97 * (sorted.size() - 1))].frametime;
     out << fixed << setprecision(1) << result << ",";
-    // avg
+    // avg + peak
     total = 0;
     for (auto input : sorted){
       total = total + input.frametime;
-      total_cpu = total_cpu + input.cpu_load;
```
alkazar, line 98: Just matching the order of the headers to make it easier to compare and validate.

src/logging.cpp
```diff
@@ -91,21 +104,66 @@ static void writeSummary(string filename){
...
       total_gpu = total_gpu + input.gpu_load;
+      total_cpu = total_cpu + input.cpu_load;
```
alkazar, line 112: Just matching the order of the headers to make it easier to compare and validate.

PR #1290, opened 2024-04-10
fix minecraft-launcher properly

Etaash-mathamsetty:
also includes a crash fix for std::stoi


Etaash-mathamsetty:
could you please revert the last 2 commits, they are not ready yet

PR #1318, opened 2024-05-03
Add nvidia shield 2017 controller

Arias800:
This commit adds support for the Nvidia shield controller.
The controller has been natively supported in the Linux kernel since 6.5.

This commit only works for the 2017 version, the 2015 version is not compatible with Linux.

Screenshot_20240503_120058

### DIFF
--- a/src/device.cpp
+++ b/src/device.cpp
@@ -19,6 +19,7 @@ int ds5_count = 0;
 int switch_count = 0;
 int bitdo_count = 0;
 int logi_count = 0; //Logitech devices, mice & keyboards etc.
+int shield_count = 0;
 
 std::string  xbox_paths [2]{"gip","xpadneo"};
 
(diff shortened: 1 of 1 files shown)


gort818 (changes requested):
The check for "thunderstrike" should be  encapsulated in if (std::find(params.device_battery.begin(), params.device_battery.end(), "gamepad") != params.device_battery.end()){

src/device.cpp
```diff
@@ -73,7 +75,16 @@ void device_update(const struct overlay_params& params){
                 device_found = true;
                 bitdo_count += 1;
             }
-    }
+        }
+
+        //CHECK NVIDIA SHIELD DEVICES
+        if (fileName.find("thunderstrike") != std::string::npos) {
```
gort818, line 81: Move this statement into the check for "gamepad" with the rest of the devices

Arias800:
I have made the requested change.
I am sorry for the error.

gort818:
No problem,  If you can squash commits ready to merge.
Thank you!

Arias800:
I think I squashed everything correctly.
It's my first time doing this, so I hope I didn't mess anything up.

gort818:
Looks good, Thanks again!

PR #1350, opened 2024-06-11
Add an option to disable ImPlot

TheComputerGuy96:
The ImPlot dependency causes significant library size and build time increases (at least for debug builds) so add an option to disable it

This change also adds more mentions of the ImPlot dependency in the documentation (and adds a warning about missing ImPlot)


flightlessmango:
I don't know if we want to maintain a non-implot version of every graph forever

PR #1374, opened 2024-07-04
update fan conditionally

lucmann:


src/overlay.cpp
```diff
@@ -749,12 +749,12 @@ void init_cpu_stats(overlay_params& params)
 {
 #ifdef __linux__
    auto& enabled = params.enabled;
-   enabled[OVERLAY_PARAM_ENABLED_cpu_stats] = cpuStats.Init()
-                           && enabled[OVERLAY_PARAM_ENABLED_cpu_stats];
-   enabled[OVERLAY_PARAM_ENABLED_cpu_temp] = cpuStats.GetCpuFile()
-                           && enabled[OVERLAY_PARAM_ENABLED_cpu_temp];
-   enabled[OVERLAY_PARAM_ENABLED_cpu_power] = cpuStats.InitCpuPowerData()
-                           && enabled[OVERLAY_PARAM_ENABLED_cpu_power];
+   enabled[OVERLAY_PARAM_ENABLED_cpu_stats] = enabled[OVERLAY_PARAM_ENABLED_cpu_stats]
```
flightlessmango, line 753: > "Failed to load NVML"
Since you're using llvmpipe, nvml should never try to initialize.
But that needs to be fixed somewhere else.
I'm thinking it might be unexpected behavior if gpu/cpu are not initialized.
Because regardless if gpu temp for example is turned off in the config, it's being turned on again while logging

PR #1398, opened 2024-08-08
param: cellpadding_x

Grief:
This allows take make the overlay smaller than it was possible:
image

### DIFF
--- a/src/overlay_params.cpp
+++ b/src/overlay_params.cpp
@@ -464,6 +464,7 @@ parse_fps_metrics(const char *str){
 #define parse_alpha(s) parse_float(s)
 #define parse_permit_upload(s) parse_unsigned(s)
 #define parse_no_small_font(s) parse_unsigned(s) != 0
+#define parse_cellpadding_x(s) parse_float(s)
 #define parse_cellpadding_y(s) parse_float(s)
 #define parse_table_columns(s) parse_unsigned(s)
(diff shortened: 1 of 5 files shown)


flightlessmango:
What does this look like in horizontal and what happens with the recording dot?
Is `cellpadding_x=1` the old default?

Grief:
@flightlessmango it works the same way, allowing to make the layout more tight
image

or loose
image

the default is `cellpadding_x=0` or not defined
image

flightlessmango:
I guess my question is if you don't set `cellpadding_x`, will it have the old spacing?
And what value is the old spacing in `cellpadding_x` ?

src/overlay_params.cpp
```diff
@@ -917,15 +919,17 @@ parse_overlay_config(struct overlay_params *params,
 
    //increase hud width if io read and write
    if (!params->width && !params->enabled[OVERLAY_PARAM_ENABLED_horizontal]) {
-      params->width = params->font_size * params->font_scale * params->table_columns * 4.6;
+      const auto real_size = params->font_size * params->font_scale;
+
+      params->width = static_cast<uint>(((1.0 + params->cellpadding_x * 2.0) * 4.6) * real_size * params->table_columns);
```
flightlessmango, line 924: mingw compiler doesn't recognize `uint`, casting as `unsigned int` should be more portable

PR #1456, opened 2024-10-07
logging: fix double-logging of avg fps

AndreFGard:
the average fps was being logged twice in the summary: once alongside other frame percentage metrics (97%, .01%, ...) and once alongside every other metric. 

I've removed the first occurence and left the latter, since it uses precision(1). The pictures illustrate the before fix and after fix. The after fix works as expected:
image
image

### DIFF
--- a/src/logging.cpp
+++ b/src/logging.cpp
@@ -104,7 +104,7 @@ static void writeSummary(string filename){
       fps_values.push_back(data.fps);
 
     std::unique_ptr<fpsMetrics> fpsmetrics;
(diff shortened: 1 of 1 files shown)


AndreFGard:
This caused every other cell after the Average FPS (column D) to shift one column to the right, resulting in more columns than expected

flightlessmango:
This looks good, I'm just trying to figure out where the other avg value comes from

flightlessmango:
I still haven't found it. Maybe you can help? it would be great to get this fixed

PR #1464, opened 2024-10-18
add intel gpu usage support via fdinfo and fix fdinfo code a little

17314642:
I didn't really dive into details, but for some reason DETECT_OS_UNIX is never defined when I'm compiling on either NixOS or when using flatpak-builder to test mangohud for steam, so I deleted that for now.

Next comes  "drm-engine-gpu" bit. "gpu" is only for msm gpus, for other gpu drivers we should use different names. That's why I added a "get_drm_engine_type" function to retrieve load type name based on gpu driver. For now I added only amdgpu, i915 and msm because they were already referenced in the project.

Last thing is the method of calculating load. Biggest change here is waiting for at least METRICS_UPDATE_PERIOD_MS (500ms) before trying to update usage percentage. On my pc old method was not working correctly, it either was always showing 100% or always showing 0% with spikes to 100%.

Can't wait for any feedback. Hopefully this commit will finally fix intel gpu usage and I can finally play without having my phone below my monitor with ssh and intel_gpu_top on the phone's screen.

### DIFF
--- a/src/gpu_fdinfo.cpp
+++ b/src/gpu_fdinfo.cpp
@@ -17,6 +17,7 @@ std::string GPU_fdinfo::get_drm_engine_type() {
 }
 
 void GPU_fdinfo::find_fd() {
+#ifdef __linux__
     DIR* dir = opendir("/proc/self/fdinfo");
     if (!dir) {
         perror("Failed to open directory");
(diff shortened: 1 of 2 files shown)


flightlessmango:
This looks great! can you just squash the "merge branch" commit?

17314642:
Done.

PR #1484, opened 2024-11-18
gpu_fdinfo: add vram usage (amd, intel) and power usage via hwmon (only intel)

17314642:
I have tested that vram usage works both on my machine (intel arc a770) and on my laptop with integrated raven2 amd gpu (with changing a line to fdinfo [here](https://github.com/flightlessmango/MangoHud/blob/a109f0baf5b5477fb1b4c3ccdfbac2284077caec/src/gpu.h#L36)). Qualcomm have not implemented vram usage via fdinfo, so no vram for msm for now :).

One thing which I'm concerned about is that power usage might not work on older intel integrated gpus. The only test subject that I have is an i5-2500M and it doesn't have any gpu sensors.

### DIFF
--- a/src/gpu_fdinfo.cpp
+++ b/src/gpu_fdinfo.cpp
@@ -168,12 +168,15 @@ void GPU_fdinfo::get_load() {
         static uint64_t previous_gpu_time, previous_time, now, gpu_time_now;
         static float power_usage_now, previous_power_usage;
 
-        gpu_time_now = get_gpu_time();
-        power_usage_now = get_power_usage();
         now = os_time_get_nano();
 
-        if (gpu_time_now > previous_gpu_time &&
-            now - previous_time > METRICS_UPDATE_PERIOD_MS * 1'000'000) {
+        if (now - previous_time < METRICS_UPDATE_PERIOD_MS * 1'000'000)
+            continue;
+
+        gpu_time_now = get_gpu_time();
+        power_usage_now = get_power_usage();
+
+        if (gpu_time_now > previous_gpu_time) {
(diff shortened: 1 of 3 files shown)


src/gpu_fdinfo.cpp
```diff
@@ -168,12 +168,15 @@ void GPU_fdinfo::get_load() {
         static uint64_t previous_gpu_time, previous_time, now, gpu_time_now;
         static float power_usage_now, previous_power_usage;
 
+        now = os_time_get_nano();
+
+        if (now - previous_time < METRICS_UPDATE_PERIOD_MS * 1'000'000)
```
flightlessmango, line 173: I believe you're right, we don't need to check this often, but I'm not sure this is sufficient.
We might be consuming a full thread like this and how it was before?

17314642: In both cases thread is using 100% of CPU core. The only way to fix it is to sleep like this in `nvidia.cpp`
https://github.com/flightlessmango/MangoHud/blob/afaa7d1b7b9212e8938e70c2ac223789832c96c6/src/nvidia.cpp#L227

image

17314642: Do you want me to use sleep and if yes, should I use METRICS_POLLING_PERIOD_MS (25ms) or METRICS_UPDATE_PERIOD_MS (500ms)?

flightlessmango: 500ms is probably fine for this

17314642: added sleep

PR #1498, opened 2024-12-05
mangoapp: pull engine from gamescope

pac85:
Gamescope side:  https://github.com/ValveSoftware/gamescope/pull/1653 

With that patch gamescope will use it's layer to pull `engineName` from `VkApplicationInfo` and send it to mangoapp.

This change will make it use that information, this allows to have the engineName displayed in a simpler way without MangoHud's layer.


src/app/main.cpp
```diff
@@ -149,6 +149,29 @@ static void gamescope_frametime(uint64_t app_frametime_ns, uint64_t latency_ns){
         HUDElements.gamescope_debug_latency.erase(HUDElements.gamescope_debug_latency.begin());
 }
 
+static EngineTypes engine_type_from_str(std::string engineName) {
```
soerengrunewald: Maybe i am wrong, but SSO by default caps at 16 characters. So should not allocate under normal circumstances. Non the less, we actually don't need a full `std::string` either.
```suggestion
static EngineTypes engine_type_from_str(std::string_view engineName) {
```

PR #1499, opened 2024-12-07
gpu_fdinfo bug fixes, almost full support for Intel is finally here

### LINKED ISSUE #1082: MangoHud Intel GPU metrics
This issue is for tracking the progress and issues with Intel GPU metrics. dGPU and iGPU.

### Current state
- `intel_gpu_top` has to be installed and it has to be setcap'd  e.g `sudo setcap cap_perfmon=+ep /usr/bin/intel_gpu_top`
- We are set up to get GPU load at least from `intel_gpu_top`
- MangoHud is only able to run `intel_gpu_top` unhindered outside of proton/runtime
- In order for us to run `intel_gpu_top` you need to have flatpak installed and set sniper runtime to beta

### LINKED ISSUE #1454: Playing Need for speed most wanted 2005 (32 bit) does not display GPU usage. It is always at 0%.
**Describe the bug**
When playing nfs mw, frame rate works but GPU usage is always at 0%. I use mangohud with bottes and downloaded mangohud from the gnome store. No other games to test unfortunately. `flatpak override --filesystem=xdg-config/MangoHud:ro` command was applied. 

- Fedora Silverblue 40/Project Bluefin stable branch
- MangoHud flathub 0.7.2
- Intel Arc Graphics that comes with intel core ultra 7 155h laptop SoC 

**To Reproduce**
1. Ensure "Monitor Performance" in bottles is toggled
2. Run NFSMW 2005 executable (speed.exe) 
3. Overlay will show GPU usage as 0%

**Expected behavior**
GPU usage needs to be displayed accurately.

**Screenshots**
Screenshot from 2024-10-05 15-51-55

17314642:
Resolves #1082
Resolves #1454

First I would like to thank you these guys for testing patches: @PerAstraAdDeum, @nokia8801 and @retrixe for helping in development.

In short, this pr fixes: 

1. wrong power usage for some people (https://github.com/flightlessmango/MangoHud/issues/1082#issuecomment-2513772813)
2. wrong gpu usage in some cases
3. vram support for integrated gpus in xe (forgot about those :crying_cat_face:)
4. 1 million watt gpu power at start (which confused one user when testing)
5. some info messages.
6. gpu clock support for i915 and xe
7. fan speed support (tested only on i915 and xe, should have linux 6.12+)
8. generic hwmon support (tested only on i915 and xe)
9. throttling support (only i915 and xe)
10. temperature support (only i915, should have linux 6.13+)
11. shows all gpus by default, otherwise user can choose which gpus he wants to see by using `pci_dev` or `gpu_list`

To summarize: pretty much every vital metric is available now for intel.

struct gpu_metrics:
  - [x] int load;
  - [X] int temp; (**only i915, no xe support by intel yet**)
  - [ ] int junction_temp {-1}; (**isn't exposed by intel**)
  - [ ] int memory_temp {-1}; (**isn't exposed by intel**)
  - [X] float memoryUsed;
  - [ ] float memoryTotal; (**isn't exposed by intel**)
  - [ ] int MemClock; (**isn't exposed by intel**)
  - [X] int CoreClock;
  - [X] float powerUsage;
  - [ ] float apu_cpu_power;
  - [ ] int apu_cpu_temp;
  - [X] bool is_power_throttled;
  - [X] bool is_current_throttled;
  - [X] bool is_temp_throttled;
  - [X] bool is_other_throttled;
  - [ ] float gtt_used; (**not used by mangohud**)
  - [X] int fan_speed; (**only i915, no xe support by intel yet**)
  - [X] int voltage;
  
  
Example metrics (using linux kernel 6.13): 
image

### DIFF
--- a/src/gpu.cpp
+++ b/src/gpu.cpp
@@ -1,4 +1,5 @@
 #include "gpu.h"
+#include <cstdint>
 #include <inttypes.h>
 #include <memory>
 #include <functional>
@@ -22,16 +23,17 @@ GPUS::GPUS(overlay_params* params) : params(params) {
     std::vector<std::string> gpu_entries;
 
     for (const auto& entry : fs::directory_iterator("/sys/class/drm")) {
-        if (entry.is_directory()) {
-            std::string node_name = entry.path().filename().string();
+        if (!entry.is_directory())
+            continue;
 
-            // Check if the directory is a render node (e.g., renderD128, renderD129, etc.)
-            if (node_name.find("renderD") == 0 && node_name.length() > 7) {
-                // Ensure the rest of the string after "renderD" is numeric
-                std::string render_number = node_name.substr(7);
-                if (std::all_of(render_number.begin(), render_number.end(), ::isdigit)) {
-                    gpu_entries.push_back(node_name);  // Store the render entry
-                }
+        std::string node_name = entry.path().filename().string();
+
+        // Check if the directory is a render node (e.g., renderD128, renderD129, etc.)
+        if (node_name.find("renderD") == 0 && node_name.length() > 7) {
+            // Ensure the rest of the string after "renderD" is numeric
+            std::string render_number = node_name.substr(7);
+            if (std::all_of(render_number.begin(), render_number.end(), ::isdigit)) {
+                gpu_entries.push_back(node_name);  // Store the render entry
             }
         }
     }
@@ -44,6 +46,8 @@ GPUS::GPUS(overlay_params* params) : params(params) {
     });
 
     // Now process the sorted GPU entries
+    uint8_t idx = 0, total_active = 0;
+
     for (const auto& node_name : gpu_entries) {
         std::string path = "/sys/class/drm/" + node_name;
         std::string device_address = get_pci_device_address(path);  // Store the result
@@ -56,7 +60,7 @@ GPUS::GPUS(overlay_params* params) : params(params) {
         } catch(...) {
             SPDLOG_ERROR("stoul failed on: {}", "/sys/bus/pci/devices/" + device_address + "/vendor");
         }
-        
+
         try {
             device_id = std::stoul(read_line("/sys/bus/pci/devices/" + device_address + "/device"), nullptr, 16);
         } catch (...) {
@@ -64,10 +68,40 @@ GPUS::GPUS(overlay_params* params) : params(params) {
         }
 
         std::shared_ptr<GPU> ptr = std::make_shared<GPU>(node_name, vendor_id, device_id, pci_dev);
+
+        if (params->gpu_list.size() == 1 && params->gpu_list[0] == idx++)
+            ptr->is_active = true;
+
+        if (!params->pci_dev.empty() && pci_dev == params->pci_dev)
+            ptr->is_active = true;
+
         available_gpus.emplace_back(ptr);
 
         SPDLOG_DEBUG("GPU Found: node_name: {}, vendor_id: {:x} device_id: {:x} pci_dev: {}", node_name, vendor_id, device_id, pci_dev);
+
+        if (ptr->is_active) {
+            SPDLOG_INFO("Set {} as active GPU (id={:x}:{:x} pci_dev={})", node_name, vendor_id, device_id, pci_dev);
+            total_active++;
+        }
     }
+
+    if (total_active < 2)
+        return;
+
+    for (auto& gpu : available_gpus) {
+        if (!gpu->is_active)
+            continue;
+
+        SPDLOG_WARN(
+            "You have more than 1 active GPU, check if you use both pci_dev "
+            "and gpu_list. If you use fps logging, MangoHud will log only "
+            "this GPU: name = {}, vendor = {:x}, pci_dev = {}",
+            gpu->name, gpu->vendor_id, gpu->pci_dev
+        );
+
+        break;
+    }
+
 }
 
 std::string GPU::is_i915_or_xe() {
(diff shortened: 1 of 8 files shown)


src/gpu_fdinfo.h
```diff
@@ -67,16 +67,28 @@ class GPU_fdinfo {
         : module(module)
         , pci_dev(pci_dev)
     {
-        SPDLOG_DEBUG("GPU driver is \"{}\"", module);
+        SPDLOG_INFO("GPU driver is \"{}\"", module);
```
flightlessmango, line 112: We don't want to print things every time unless it's not expected, this should be debug

src/gpu_fdinfo.h
```diff
@@ -67,16 +67,28 @@ class GPU_fdinfo {
...
             drm_memory_type = "drm-total-local0";
         } else if (module == "xe") {
             drm_engine_type = "drm-total-cycles-rcs";
             drm_memory_type = "drm-resident-vram0";
+
+            if (
+                fdinfo_data.size() > 0 &&
+                fdinfo_data[0].find(drm_memory_type) == fdinfo_data[0].end()
+            ) {
+                SPDLOG_INFO(
```
flightlessmango, line 128: Should also be debug

src/gpu.cpp
```diff
@@ -66,7 +66,7 @@ GPUS::GPUS(overlay_params* params) : params(params) {
         std::shared_ptr<GPU> ptr = std::make_shared<GPU>(node_name, vendor_id, device_id, pci_dev);
         available_gpus.emplace_back(ptr);
 
-        SPDLOG_DEBUG("GPU Found: node_name: {}, vendor_id: {:x} device_id: {:x} pci_dev: {}", node_name, vendor_id, device_id, pci_dev);
```
flightlessmango, line 69: should be debug

src/gpu.cpp
```diff
@@ -66,7 +66,7 @@ GPUS::GPUS(overlay_params* params) : params(params) {
...
+        SPDLOG_INFO("GPU Found: node_name: {}, vendor_id: {:x} device_id: {:x} pci_dev: {}", node_name, vendor_id, device_id, pci_dev);
```
flightlessmango, line 80: should be debug

src/gpu_fdinfo.h
17314642: Gpu driver is also printen only once

src/gpu.cpp
```diff
@@ -153,7 +153,7 @@ void GPUS::find_active_gpu() {
                 for (const auto& gpu : available_gpus) {
                     if (gpu->pci_dev == drm_pdev) {
                         gpu->is_active = true;
-                        SPDLOG_DEBUG("Active GPU Found: node_name: {}, pci_dev: {}", gpu->name, gpu->pci_dev);
```
flightlessmango, line 156: should be debug

src/gpu.cpp
```diff
@@ -173,7 +173,7 @@ void GPUS::find_active_gpu() {
                 for (auto& pid : gpu->nvidia_pids()) {
                     if (pid == getpid()) {
                         gpu->is_active = true;
-                        SPDLOG_DEBUG("Active GPU Found: node_name: {}, pci_dev: {}", gpu->name, gpu->pci_dev);
```
flightlessmango, line 176: should be debug

src/gpu_fdinfo.h
17314642: Same thing

src/gpu.cpp
17314642: Shouldn't user be able to see this easier? Considering these things get printed once per launch. Instead of asking him to change his SPDLOG_LEVEL.

src/hud_elements.cpp
```diff
@@ -250,22 +250,20 @@ void HudElements::gpu_stats(){
                 ImGui::PopFont();
             }
 
-            if (HUDElements.vendorID == 0x1002 || HUDElements.vendorID == 0x10de){
-                if (HUDElements.params->enabled[OVERLAY_PARAM_ENABLED_gpu_fan] && cpuStats.cpu_type != "APU"){
-                    ImguiNextColumnOrNewRow();
-                    right_aligned_text(text_color, HUDElements.ralign_width, "%i", gpu->metrics.fan_speed);
+            if (HUDElements.params->enabled[OVERLAY_PARAM_ENABLED_gpu_fan] && cpuStats.cpu_type != "APU"){
```
flightlessmango, line 253: `cpuStats.cpu_type` should be deprecated in favor of `gpu->is_apu()`.
Only AMD is being set as APU atm so intel needs to be hooked up there

17314642: cpu_type is completetly unused, even for amd

17314642: and do you mean it needs changing from `cpuStats.cpu_type != "APU"` to `!gpu->is_apu()`?

src/gpu.cpp
flightlessmango: The problem is that it's usually not once per launch, mangohud attaches to many random processes and spams the logs.
Unless something is out of the ordinary or something is wrong is when we want to print

src/gpu.cpp
17314642: since find_active_gpu is nuked, i dont think this is necessary

flightlessmango: You're right :+1:

src/hud_elements.cpp
17314642: done, i think?

17314642: and I don't know currently how to detect if gpu is an integrated one, so that will have to wait

flightlessmango: Yes it should be changed to `!gpu->is_apu()`

17314642: I changed it already

src/gpu_fdinfo.h
flightlessmango: As mentioned earlier, the problem is that nothing is mentioned once when launched from steam etc, it will spam the logs

src/gpu.cpp
flightlessmango: Oh it's been changed already, I'm learning I don't understand how this tool works.

17314642: yeah, it's not really intuitive, i already force pushed all the changes

Issue #1497, opened 2024-12-05
MangoHud still relies on X11 even when built with_x11=disabled and with_wayland=enabled


nokia8801:
**Describe the bug**
When built with `with_x11=disabled` and `with_wayland=enabled`, the HUD position is wrong and doesn't follow the config value, RShift_F11 doesn't change the position as none of the shortcuts work, `display-server` config option also doesn't work.

**List relevant hardware/software information**
- **OS:** Arch Linux
- **MangoHud:** 0.7.2
- **GPU:** Intel Arc A750
- **Mesa:** 24.2.7
- **Wine:** wine-tkg-git 9.22.r2.g342b3b81-327  ( TkG Staging NTsync )
- **DE/WM:** GNOME 47.2 on Wayland (Mutter)
- **DXVK:** 2.5.1
- **Kernel:** Linux 6.12.1-arch1-1

**To Reproduce**
When running a game natively over Wayland, such as Windows games using Wine with the graphics backend set to Wayland (without x11 fallback) or Minecraft with GLFW 3.4 running natively over Wayland, if we build MangoHud with `with_x11=disabled` and `with_wayland=enabled`, the HUD position will be wrong, shortcuts won't work and `display-server` will return empty.

**Expected behavior**
Shortcuts, `display-server` and `position` should work.

**Screenshots**
![image](https://github.com/user-attachments/assets/28cd010a-732b-4b7d-aa6a-79ab1c03a505)

**Additional context**
I made this issue back in April, which could be related. https://github.com/flightlessmango/MangoHud/issues/1281

I build Wine with the Wayland driver and set it as the only graphics backend. Wine is also built with and using WoW64. I also use GLFW 3.4 with Minecraft. Both Wine games and Minecraft run natively over Wayland. When the games are running without MangoHud, no `xwayland` or `mutter-x11-frames` are spawned/running/sleeping. I guess something got fixed since April and now, because now, even with MangoHud, those processes aren't spawned/running/sleeping at all. So that's exactly what I wanted. Thanks. It can be used without any X11 dependencies or calls. (Hopefully I'm correct in that last part)

But the HUD is stuck and not configurable. That's the issue.

I should also mention that I don't use mangoapp or mangohudctl, so those may affect the building process or using it as Wayland only. Here's my PKGBUILD options:
```
build() {
    arch-meson "$_pkgname" build \
        -Dmangoapp=false \
        -Dmangohudctl=false \
        -Dwith_nvml=disabled \
[...]

PR #1500, opened 2024-12-08
Don't rely on x11 when with_x11 is disabled

Etaash-mathamsetty:
Fixes (partially): #1497 
Part 1... will expand in the future to switch everything to xkb_keysym_t, seems like that's what we are forced to do


nokia8801:
That whole active GPU stuff in src/gpu.cpp got fixed with https://github.com/flightlessmango/MangoHud/pull/1499, so it is safe to remove as there is no need for it anymore. This would resolve the conflict.

Etaash-mathamsetty:
> That whole active GPU stuff in src/gpu.cpp got fixed with #1499, so it is safe to remove as there is no need for it anymore. This would resolve the conflict.

I am aware

PR #1503, opened 2024-12-09
Add OpenGL Shim

Etaash-mathamsetty:


src/gl/shim.c
```diff
@@ -36,7 +48,7 @@ static void loadMangoHud() {
 
     if (mangoHudLoaded) {
         printf("shim: Loaded mangohud library\n");
```
flightlessmango, line 50: We don't want to print to stdout from shim, use stderr

src/meson.build
```diff
@@ -222,6 +222,7 @@ mangohud_opengl_shared_lib = shared_library(
   opengl_files,
   vklayer_files,
   util_files,
+  gnu_symbol_visibility : 'hidden',
```
flightlessmango, line 225: Not sure if you or I added this, but it seems irrelevant to this PR?

Etaash-mathamsetty: it was part of your original code, I thought it was maybe used with the DEEPBIND for something

PR #1512, opened 2024-12-26
Fixes for shim

Etaash-mathamsetty:
dlsym hooking is now default

pending stuff:
- [x] Add env to disable dlsym hooking (done)
- [x] Update README with changes


src/gl/shim.c
```diff
@@ -113,16 +114,109 @@ int64_t glXSwapBuffersMscOML(void* dpy, void* drawable, int64_t target_msc, int6
...
+};
+#undef ADD_HOOK
+
+#ifndef ARRAY_SIZE
+#define ARRAY_SIZE(arr) sizeof(arr)/sizeof(arr[0])
+#endif
+
+void* dlsym(void *handle, const char *name)
+{
+    const char* dlsym_enabled = getenv("MANGOHUD_DLSYM");
```
flightlessmango, line 206: The thing is we can't know if the user set `--dlsym` in the mangohud script.
Maybe it would make more sense to just use dlsym by default and add an env for disabling it?

Etaash-mathamsetty: that's what it does lol

flightlessmango: So then `--dlsym` does nothing. I guess we could just retire that?

Etaash-mathamsetty: yeah it's been removed from the mangohud script (I didn't update the README in this PR, I probably should)

PR #1514, opened 2024-12-29
refactor memory.cpp

17314642:
looks a lot cleaner now and start using `MemTotal - MemAvailable` since everyone but htop use this now

### DIFF
--- a/src/memory.h
+++ b/src/memory.h
@@ -2,28 +2,12 @@
 #ifndef MANGOHUD_MEMORY_H
 #define MANGOHUD_MEMORY_H
 
-#include <stdio.h>
-#include <thread>
+#include <cstdint>
 
-extern float memused, memmax, swapused, swapmax, rss;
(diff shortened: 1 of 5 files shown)


src/memory.h
```diff
@@ -2,28 +2,12 @@
...
 
-struct memory_information {
-  /* memory information in kilobytes */
-  unsigned long long mem, memwithbuffers, memeasyfree, memfree, memmax,
-      memdirty;
-  unsigned long long swap, swapfree, swapmax;
-  unsigned long long bufmem, buffers, cached;
-};
-
-struct process_mem
```
flightlessmango, line 18: Why couldn't we still have a struct for this?

17314642: I was confused as to why process memory has a struct, but not ram and just decided to delete it for consistency

17314642:
Rebased and added gamescope support for proc_mem and io_read/write

PR #1532, opened 2025-01-18
[blacklist] Add Plutonium launcher

nokia8801:
Adds Plutonium Launcher to the blacklist which is a custom client for dedicated servers on old Call of Duty games.


nokia8801:
Avoids this:

image

Games launched still display the overlay, as I did not add `plutonium-bootstrapper-win32.exe` to the blacklist, which is responsible for launching the games.

PR #1546, opened 2025-01-30
add mentions of 32-bit version and fix opensuse url

17314642:


README.md
```diff
@@ -192,21 +194,23 @@ sudo pacman -Syy
 
 If you are using Debian 11 (Bullseye) or later, Ubuntu 21.10 (Impish) or later, or distro derived from them, to install the [MangoHud](https://tracker.debian.org/pkg/mangohud) package, execute:
 
+NOTE: debian-based distributions have no 32-bit support, so if you need it, either compile yourself, or create a request on debian bug tracker.
```
smcv, line 197: This is not true: Debian has a 32-bit mangohud available. A 32-bit apt source is not configured by default, but if you can run 32-bit GL or Vulkan applications, you must have already enabled 32-bit (`sudo dpkg --add-architecture i386 && apt update`) and if you can't run 32-bit GL or Vulkan applications then 32-bit mangohud isn't useful.

*Ubuntu* does not have a 32-bit mangohud available.

I've opened #1547 to clarify this.

17314642: You're right. Should I just change it to something like:

"Debian has 32-bit support, but other debian-based distributions may not, check your own distro's repository"

smcv: I tried to clarify this in https://github.com/flightlessmango/MangoHud/pull/1547. Does what's in my PR look good to you? I tried to be as specific as I could. If you like my proposed wording, you could drop the Debian part from this PR and leave that for #1547.

Or, if you don't like my proposed wording, please suggest something different, either here or on #1547 or in a separate PR.

As a general principle, if you're making more than one orthogonal change at the same time, it's often good for each part to be a separate commit (or even a separate PR) so that each part can be accepted, rejected or discussed further without affecting the status of the other parts.

> Debian has 32-bit support, but other debian-based distributions may not, check with your own distro's repository

With Ubuntu being rather popular and lacking `mangohud:i386`, it seems worthwhile to explicitly say that it's known not to be available on Ubuntu.

I think "32-bit support" is a bit unclear - Ubuntu does support running 32-bit games, and it does have 32-bit graphics drivers etc., it just doesn't have 32-bit versions of less critical packages (which unfortunately includes Mangohud).

flightlessmango:
merged #1547

17314642:
removed mention about debian

README.md
17314642: yeah, it was bad wording on my part, i wanted to imply that only 32-bit mangohud is unavailable, but 32-bit apps in general. anyway, i removed that mention.

and thanks for mention about creating separate prs, even though in this case i thought that it's such a small change, that no separate commits or prs were necessary

PR #1601, opened 2025-03-03
hud_elements.cpp: fix imgui GetWindowContentRegionWidth() -> GetContentRegionAvail().x

Austrum-lab:
https://github.com/ocornut/imgui/blob/master/imgui.cpp#L682

GetWindowContentRegionWidth() was removed from imgui

### DIFF
--- a/src/hud_elements.cpp
+++ b/src/hud_elements.cpp
@@ -407,7 +407,7 @@ void HudElements::core_load(){
             width = 150;
             height = HUDElements.params->font_size;
         } else {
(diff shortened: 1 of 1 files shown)


flightlessmango:
reverted this PR as `ImGui::GetWindowContentRegionWidth();` and `ImGui::GetContentRegionAvail().x` is not the same.
 In my test
`ImGui::GetContentRegionAvail().x` = 102
`ImGui::GetWindowContentRegionWidth()` = 321

PR #1617, opened 2025-03-16
Add 'horizontal separator color' parameter

Cuberootex:
Adds parameter to customize the color of the horizontal separator when using the horizontal layout, since by default it is hardcoded to be the same as the VRAM's color. Additionally this change makes it respect the alpha parameter.

### DIFF
--- a/data/MangoHud.conf
+++ b/data/MangoHud.conf
@@ -319,6 +319,7 @@ text_outline
 # wine_color=EB5B5B
 # battery_color=FF9078
 # network_color=E07B85
+# horizontal_separator_color=AD64C1
 
 ### Specify GPU with PCI bus ID
 ### Set to 'domain:bus:slot.function'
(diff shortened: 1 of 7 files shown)


src/overlay_params.cpp
```diff
@@ -781,6 +782,7 @@ static void set_param_defaults(struct overlay_params *params){
    params->media_player_name = "";
    params->font_scale = 1.0f;
    params->wine_color = 0xeb5b5b;
+   params->horizontal_separator_color = 0xa2a2a2;
```
flightlessmango, line 785: the default color should still be the same as vram

data/MangoHud.conf
```diff
@@ -319,6 +319,7 @@ text_outline
 # wine_color=EB5B5B
 # battery_color=FF9078
 # network_color=E07B85
+# horizontal_separator_color=A2A2A2
```
flightlessmango, line 322: should be vram color here as well.
Also squish this into one commit

PR #1622, opened 2025-03-19
Show GPU efficiency

gnusenpai:
Adds a HUD element that displays GPU efficiency in frames per joule (F/J), also sometimes referred to frames per second per watt (FPS/W).

Useful for getting immediate feedback when GPU undervolting and power limiting.

Optionally enable flip_efficiency to show joules per frame (J/F) instead.

If there is interest, I can work on a similar feature for CPU efficiency.

vkcube

### DIFF
--- a/README.md
+++ b/README.md
@@ -406,6 +406,8 @@ Parameters that are enabled by default have to be explicitly disabled. These (cu
 | `gpu_name`                         | Display GPU name from pci.ids                                                         |
 | `gpu_voltage`                      | Display GPU voltage (only works on AMD GPUs)                                          |
 | `gpu_list`                         | List GPUs to display `gpu_list=0,1`                                                   |
+| `gpu_efficiency`                   | Display GPU efficiency in frames per joule                                            |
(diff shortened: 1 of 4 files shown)


flightlessmango:
would it make sense to have an option for frames per watt as well?
seems like a more approachable value to me.
Also the indicator of the max power consumption is nice, can you make a PR for that as well?

gnusenpai:
> would it make sense to have an option for frames per watt as well? seems like a more approachable value to me.

AFAIU, frames per watt (F/W) specifically isn't mathematically correct. Frames per second per watt (FPS/W) is more correct. While intuitive, it is technically redundant since it is the same as frames per second per joule per second (F/s / J/s) which is a convoluted unit. You can cancel out the seconds and simplify it to frames per joule (F/J).

To put it simply, either FPS/W or F/J are valid and the same.

I can make an option for it, but maybe it is a bit silly just to save 2 characters...

> Also the indicator of the max power consumption is nice, can you make a PR for that as well?

Sure! Although my implementation currently only supports NVIDIA because that is all I can test.

PR #1623, opened 2025-03-20
Show GPU power limit

gnusenpai:
As requested in #1622, this modifies the GPU power HUD element to include the current power limit.

Currently, the power limit label is not optional. Should it be?

This is currently marked as draft since it currently only supports NVIDIA.
AMD/Intel help wanted!

### DIFF
--- a/src/nvidia.cpp
+++ b/src/nvidia.cpp
@@ -118,9 +118,11 @@ void NVIDIA::get_instant_metrics_nvml(struct gpu_metrics *metrics) {
         }
 
         if (params->enabled[OVERLAY_PARAM_ENABLED_gpu_power] || (logger && logger->is_active())) {
-            unsigned int power;
+            unsigned int power, limit;
             nvml.nvmlDeviceGetPowerUsage(device, &power);
+            nvml.nvmlDeviceGetPowerManagementLimit(device, &limit);
             metrics->powerUsage = power / 1000;
+            metrics->powerLimit = limit / 1000;
         }
 
         if (params->enabled[OVERLAY_PARAM_ENABLED_throttling_status]) {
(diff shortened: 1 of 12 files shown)


17314642:
i sent a pr for intel support

src/nvidia.cpp
```diff
@@ -238,6 +240,7 @@ void NVIDIA::get_samples_and_copy() {
         cond_var.wait(lock, [this]() { return !paused || stop_thread; });
         GPU_UPDATE_METRIC_AVERAGE(load);
         GPU_UPDATE_METRIC_AVERAGE_FLOAT(powerUsage);
+        GPU_UPDATE_METRIC_AVERAGE_FLOAT(powerLimit);
```
flightlessmango, line 243: This should probably be `GPU_UPDATE_METRIC_MAX`

gnusenpai:
> probably yes, because hwmon interface (which is also used for power limit data) is not available at all for intel's integrated gpus for both i915 and xe drivers

We could either have a configurable parameter or some simple check whether to display power limit or not.
Something like: if powerLimit == 0, then don't print it to the element.

Which approach is better?

17314642:
probably a config option, explicit is better than implicit

gnusenpai:
Should work on AMD now, tested with a 6700XT and a Steam Deck.

I noticed Steam Deck has two power limits in hwmon, something about slowPPT and fastPPT. Both were the same for me (15W). There may need to be special considerations for other similar configurations.

flightlessmango:
This doesn't need to be multiple commits. It will make it difficult to bisect later on if needed

gnusenpai:
Just remembered: should this be guarded like power_usage is?
https://github.com/flightlessmango/MangoHud/blob/e906c6b03435678691e2f12340e318f62573fdb0/src/amdgpu.cpp#L288-L296

17314642:
if test fail to compile, then yes

17314642:
actually no, since github successfully compiled mangohud

flightlessmango:
> Just remembered: should this be guarded like power_usage is?

Yeah if we're not getting power1_average from sysfs, there's no point in getting power limit either. I assume there's a risk that node might be causing lag as well

17314642:
ahahah, i thought by guarding he meant that #ifdef statement

gnusenpai:
thanks to @17314642 for the intel support and testing

PR #1627, opened 2025-03-25
add core_type option

17314642:
Shows P and E cores on intel and cortex model on arm. This option supports intel and most popular arm processors (defined in cpu.h at line 196). To use this option, core_load must be also enabled.

### Intel example
core_type_intel
*There is no E-cores because I have an i5-12400F*

### ARM example
core_type_arm


flightlessmango:
This only needs to be one commit

17314642:
is it okay that those commits are not related? that's why i separated them

PR #1629, opened 2025-03-26
Show CPU efficiency

gnusenpai:
CPU analog of #1622.

Adds a HUD element that displays CPU efficiency in frames per joule (F/J).

I would be interested in adding a CPU power limit indicator like in #1623, but AFAIK, CPUs don't seem to report that without poking at the MSRs and such.

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -358,6 +358,7 @@ Parameters that are enabled by default have to be explicitly disabled. These (cu
 | `cpu_power`<br>`gpu_power`         | Display CPU/GPU draw in watts                                                         |
 | `cpu_temp`<br>`gpu_temp`<br>`gpu_junction_temp`<br>`gpu_mem_temp`           | Display current CPU/GPU temperature                                                  |
 | `cpu_text`<br>`gpu_text`           | Override CPU and GPU text. `gpu_text` is a list in case of multiple GPUs              |
+| `cpu_efficiency`                   | Display CPU efficiency in frames per joule                                            |
 | `custom_text_center`               | Display a custom text centered useful for a header e.g `custom_text_center=FlightLessMango Benchmarks` |
 | `custom_text`                      | Display a custom text e.g `custom_text=Fsync enabled`                                 |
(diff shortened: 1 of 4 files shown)


gnusenpai:
Maybe `flip_efficiency` should be moved since it now applies to both CPU and GPU?
https://github.com/flightlessmango/MangoHud/blob/f8a36e96107fa875806b7a87d79a0d3bca56c7b4/README.md?plain=1#L409-L411
https://github.com/flightlessmango/MangoHud/blob/f8a36e96107fa875806b7a87d79a0d3bca56c7b4/data/MangoHud.conf#L96-L98

flightlessmango:
> Maybe flip_efficiency should be moved since it now applies to both CPU and GPU?

Moved where/how? But probably yes?

flightlessmango:
I see what you mean. Can't really think of a good place to put it

flightlessmango:
This might be interesting to have for devices like steam deck where you calculate based on the battery wattage

gnusenpai:
Something like `system_efficiency`?

flightlessmango:
Yeah I think that sounds good

gnusenpai:
I've moved `flip_efficiency` to where I think it makes sense. It appears the variables in the README *were* sorted alphabetically at some point, but it no longer is so I just put it at the bottom.

Do you have a development workflow for testing on the Steam Deck? I made it happen when making the power limit PR, but it was a complete mess.  Kinda need that sorted to work on `system_efficiency`.

flightlessmango:
> Do you have a development workflow for testing on the Steam Deck?

No not really

axet:
Because we have GPU measured in WATT would be nice to convert JOULES to WATTs. W=J/s

flightlessmango:
This looks good. Can you fix the conflict just?

gnusenpai:
@flightlessmango looks like master has some variables that didn't get renamed (`font1` -> `font_small`) and is causing the build to fail, but what I touch should be fixed now.

flightlessmango:
It's been reverted, if you force push again I assume tests will run again

PR #1630, opened 2025-03-27
add procmem for legacy_layout and make xnvctrl work without nvml

17314642:
i tested procmem with:

1) just procmem
2) procmem + horizontal
3) procmem + hud_compact
4) procmem + horizontal + hud_compact

and same as above but with procmem + procmem_shared + procmem_virt

everything looked correct

### DIFF
--- a/src/nvidia.cpp
+++ b/src/nvidia.cpp
@@ -66,15 +66,14 @@ NVIDIA::NVIDIA(const char* pciBusId) {
     }
 
 #endif
-#ifdef HAVE_NVML
-    if (nvml_available || nvctrl_available){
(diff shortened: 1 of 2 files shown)


src/nvidia.cpp
```diff
@@ -66,15 +66,14 @@ NVIDIA::NVIDIA(const char* pciBusId) {
     }
 
 #endif
-#ifdef HAVE_NVML
```
flightlessmango, line 69: does this compile with nvml disabled?

17314642: yes

Issue #1237, opened 2024-02-16
Show RAM and VRAM usage of the application itself


shmerl:
Currently mangohud shows system RAM / VRAM utilization. It would be nice if possible to see how much memory the actual program that runs with the HUD uses.

shmerl:
Any news on this? Showing total is not a very easy way to understand how application itself affects VRAM.

Lassebq:
You can use `procmem` for process RAM usage. Nothing for vram sadly. Now that you mention it I would like to see process vram too.

shmerl:
Thanks for the pointer about procmem

shmerl:
I must be missing something. I tired adding `...,procmem,...` to the list of options I'm using, but it's not showing me anything about it.

flightlessmango:
It appears procmem is only usable with `legacy_layout=0`, this should probably be changed tho

shmerl:
That helped indeed. What does it do exactly? It's not documented in readme besides being mentioned for `exec`.

flightlessmango:
It only displays the elements that are in the config and it allows you to re-order the elements

shmerl:
That should probably be a default.

shmerl:
Looks like a bunch of options aren't working with `legacy_layout=0`?

flightlessmango:
For example to see gpu information you need to add `gpu_stats`

shmerl:
That worked, thanks. Is it possible to read total VRAM of the process in general?

Lassebq:
nvtop does have per process ram. Should be possible.

flightlessmango:
Yeah there's no issues implementing it, just time to write it

shmerl:
Looks like this works now, with `proc_vram`. So I'll close it.

shmerl:
Actually, for me it just reports 0 all the time for proc_vram, so something isn't working right.

shmerl:
Let's open a new issue about it not working properly

shmerl:
Opened bug #1879 instead.

PR #1631, opened 2025-03-27
add proc_vram option

17314642:
Closes issue #1237

this option shows vram usage by the current process

#### NVIDIA
Screenshot from 2025-03-27 04-37-47

#### AMD
Screenshot from 2025-03-27 07-18-25

#### Intel, panfrost, msm
they already read vram this way, i just moved readings from vram to pvram


flightlessmango:
Looks good to me, just needs to resolve conflict

17314642:
done

nokia8801:
VRAM is now empty on Intel Xe discrete GPU.

image

> Intel, panfrost, msm
> they already read vram this way, i just moved readings from vram to pvram

So it's intentional? I'm guessing from now on we use `proc_vram` instead of `vram`?

17314642:
Yes, it's intentional. Technically no Intel or msm or panfrost can show total vram yet, only process vram.

If it still showed that as VRAM and not PVRAM, that would mean that mangohud is hiding another 0.5-1.5 GB system VRAM usage from user, essentially showing incorrect data, where user might think that Intel consumes less vram than amd or nvidia, while in reality it's not the case

PR #1651, opened 2025-04-13
cpu: attempt to use RAPL on non-Intel CPUs

philipl:
Despite the strong association with Intel, RAPL is not actually Intel specific and works for AMD Zen processors. In fact, it seems like new processors do not report power consumption in any way except RAPL.

So, simply attempt to use RAPL for any CPU.

Tested on a 9950x3D


philipl:
Done. (I did a quick check in the kernel code and all Zen microarchitecture CPUs are supported)

PR #1655, opened 2025-04-19
nvidia: fix crash because of uninitialized boolean

17314642:
primitive types are uninitialized by default, fix that

### DIFF
--- a/src/nvidia.h
+++ b/src/nvidia.h
@@ -11,8 +11,10 @@
 class NVIDIA {
     public:
         std::shared_ptr<Throttling> throttling;
-        bool nvml_available;
-        
+
+        bool nvml_available = false;
+        bool nvctrl_available = false;
+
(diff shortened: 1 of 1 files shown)


flightlessmango:
Should do the same for `bool nvctrl_available;`

17314642:
done

src/nvidia.h
```diff
@@ -66,9 +68,6 @@ class NVIDIA {
         nvmlDevice_t device;
         void get_instant_metrics_nvml(struct gpu_metrics *metrics);
 #endif
-        bool nvctrl_available;
-        bool failed;
```
flightlessmango, line 70: I assume failed was never used?

17314642: Yes, its a private class variable which is not referenced anywhere

Issue #1607, opened 2025-03-10
GPU power draw shows wrong numbers with 32bit applications


xpander69:
**Describe the bug**
32bit games/applications show powerdraw in 100k numbers or more.

**List relevant hardware/software information**
- Arch Linux
- MangoHud version: v0.8.1-4-gcd7c6cb
- GPU: nvidia RTX 3080, 570.124.04 drivers

**To Reproduce**
1. open mangohud with any 32bit application like glxgears32 for example for quick test

**Screenshots**
32bit glxgears:

![Image](https://github.com/user-attachments/assets/3c82209d-b8a0-4571-8910-d12c6ff9b999)

64bit glxgears:

![Image](https://github.com/user-attachments/assets/bc0962b4-b9f0-4923-8f2e-632df108122c)

Linerd4:
This is Alice: Madness Returns, a 2011 year-old game, played on Steam with Proton.

![Image](https://github.com/user-attachments/assets/7a1f9891-26e4-498c-a9a0-68809d1a35c8)

I'm using Arch Linux, with Wayland + Sway, and with the following versions of Mangohud downloaded from the official Arch repositories:
extra/mangohud 0.8.1-1 [installed]
multilib/lib32-mangohud 0.8.1-1 [installed]

The followings are my PC's specs:
Nvidia RTX 3050 6 gb
Intel i7-7700
16 gb DDR4 2400 MHz

Games like Minecraft (native), Hollow Knight (native), Hades (Proton), Resident Evil 3 Remake (Proton), Severed Steel (GOG-Lutris-Proton) played all fine, maybe it could be an issue related with the 32-bit build.

Have I missed some steps to fix the random wattage values from the GPU ?

JStrategic:
Same issue on Fedora Wayland
Nvidia RTX 3080 12GB (Driver: 570.133.07)
Mangohud version: 0.8.1-2

xpander69:
it might be 3000 series GPUs only issue. I seen people on 4000 series don't have this issue. So it could be some strange nvidia bug, but then again nvidia-smi reports all correctly

flightlessmango:
```
diff --git a/src/nvidia.cpp b/src/nvidia.cpp
index b983656..e620387 100644
--- a/src/nvidia.cpp
+++ b/src/nvidia.cpp
@@ -118,9 +118,16 @@ void NVIDIA::get_instant_metrics_nvml(struct gpu_metrics *metrics) {

[...]

PR #1656, opened 2025-04-19
nvidia: fix power usage and limit reporting by ensuring float division

Zebra2711:
my game's gpu power usage is below 0.09W, and magohud sometimes displays the wrong power usage with a large value

<div align="center">

</div>

### DIFF
--- a/src/nvidia.cpp
+++ b/src/nvidia.cpp
@@ -120,8 +120,8 @@ void NVIDIA::get_instant_metrics_nvml(struct gpu_metrics *metrics) {
             unsigned int power, limit;
             nvml.nvmlDeviceGetPowerUsage(device, &power);
             nvml.nvmlDeviceGetPowerManagementLimit(device, &limit);
-            metrics->powerUsage = power / 1000;
-            metrics->powerLimit = limit / 1000;
+            metrics->powerUsage = power / 1000.0f;
+            metrics->powerLimit = limit / 1000.0f;
         }
 
         if (params->enabled[OVERLAY_PARAM_ENABLED_throttling_status]) {


flightlessmango:
Does this only occur with 32bit apps like this issue https://github.com/flightlessmango/MangoHud/issues/1607 ?
If it is the same, we believe it's a bug in the nvidia driver.
0.08 watt while producing 60 frames also sounds bugged

Zebra2711:
> Does this only occur with 32bit apps like this issue #1607 ? If it is the same, we believe it's a bug in the nvidia driver. 0.08 watt while producing 60 frames also sounds bugged

I can't reproduce that in NVIDIA driver version 575.51.02, so I think it got fixed? Now it always shows 0.0W

17314642:
> > Does this only occur with 32bit apps like this issue #1607 ? If it is the same, we believe it's a bug in the nvidia driver. 0.08 watt while producing 60 frames also sounds bugged
> 
> I can't reproduce that in NVIDIA driver version 575.51.02, so I think it got fixed? Now it always shows 0.0W

This user still has 0.08W, same driver version

https://github.com/flightlessmango/MangoHud/issues/1607#issuecomment-2810525669

Zebra2711:
> > I can't reproduce that in NVIDIA driver version 575.51.02, so I think it got fixed? Now it always shows 0.0W
> 
> This user still has 0.08W, same driver version
> 
> [#1607 (comment)](https://github.com/flightlessmango/MangoHud/issues/1607#issuecomment-2810525669)

you mean 0.008W?  So it's still not fixed, but the gpu power usage no longer reports large values like before

PR #1659, opened 2025-04-21
Add arm64 builds

Etaash-mathamsetty:
I wonder if it's better to just do a cross compile, but GitHub gives us arm runners so im not sure


flightlessmango:
The arm jobs don't seem to be working? Just waiting to be picked up

Etaash-mathamsetty:
> The arm jobs don't seem to be working? Just waiting to be picked up

yeah that's odd

PR #1661, opened 2025-04-21
Multiple segfaults fix

17314642:
Down below are all commit messages in one place.

All of these are different commits for cleaner commit history and easier understanding of the changes.

- fix x11 segfault
  - when closing an application, x11 library might get unloaded while mangohud threads are still working leading to use-after-free

- fix gpu_fdinfo segfault
  - when closing an application, GPU_fdinfo will get destroyed without joining GPU_fdinfo::main_thread() leading to use-after-free

- fix amdgpu segfault
  - when closing an application AMDGPU will get destroyed without joining AMDGPU::metrics_polling_thread() leading to use-after-free

- fix compilation without dbus
  - when compiling with -Dwith_dbus=disabled, compiler complains about unused function.

- fix nvidia segfault
  - 1\) libnvml_loader might get destroyed before nvidia thread stops working leading to use-after-free.
     - fix this by using std::shared_ptr so that libnvml_loader will be destroyed only after references to libnvml_loader is zero.
     -
     - same thing goes for libxnvctrl_loader.
     -
     - also add RTLD_NODELETE just in-case.

  - 2\) NVIDIA will get destroyed without joining NVIDIA::get_samples_and_copy() leading to use-after-free.

### DIFF
--- a/src/gpu_fdinfo.h
+++ b/src/gpu_fdinfo.h
@@ -204,6 +204,7 @@ public:
             find_xe_gt_dir();
 
         thread = std::thread(&GPU_fdinfo::main_thread, this);
+        pthread_setname_np(thread.native_handle(), "gpu_fdinfo");
     }
 
     ~GPU_fdinfo() {
--- a/src/gpu_fdinfo.h
+++ b/src/gpu_fdinfo.h
@@ -203,8 +203,13 @@ public:
         else if (module == "xe")
             find_xe_gt_dir();
 
-        std::thread thread(&GPU_fdinfo::main_thread, this);
-        thread.detach();
+        thread = std::thread(&GPU_fdinfo::main_thread, this);
+    }
+
+    ~GPU_fdinfo() {
+        stop_thread = true;
+        if (thread.joinable())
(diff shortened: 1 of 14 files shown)


src/gpu_fdinfo.h
```diff
@@ -203,8 +203,13 @@ class GPU_fdinfo {
         else if (module == "xe")
             find_xe_gt_dir();
 
-        std::thread thread(&GPU_fdinfo::main_thread, this);
-        thread.detach();
+        thread = std::thread(&GPU_fdinfo::main_thread, this);
```
flightlessmango, line 206: thread doesn't need to be detached here?

17314642: Why do you think it should?

17314642:
Also forgot to mention. Using `std::thread thread(...)` is incorrect. That's why I changed it to `thread = std::thread(...)`.

Because when using `std::thread thread(...)`, it actually creates a new local scope variable called `thread` which overshadows class' defined variable and doesn't actually assign to an already existing `thread` variable.

PR #1664, opened 2025-04-24
When setting thread names, give a hint that they are for MangoHud

smcv:
* When setting thread names, give a hint that they are for MangoHud
    
    Because MangoHud is a LD_PRELOAD module or a Vulkan layer, its threads
    run as part of some host process, typically a game. Putting a common
    prefix on the threads means they can be distinguished from threads
    created by some other library loaded into the same host process.
    
    For instance, renaming the D-Bus thread from dbus to mangohud-dbus allows
    it to be distinguished from some other library also using a thread to
    access D-Bus.
    
    Thread names are limited to 15 bytes + '\0', so some of them needed to be
    abbreviated appropriately.

### DIFF
--- a/src/amdgpu.cpp
+++ b/src/amdgpu.cpp
@@ -435,5 +435,5 @@ AMDGPU::AMDGPU(std::string pci_dev, uint32_t device_id, uint32_t vendor_id) {
 #endif
 
 	thread = std::thread(&AMDGPU::metrics_polling_thread, this);
-	pthread_setname_np(thread.native_handle(), "amdgpu");
+	pthread_setname_np(thread.native_handle(), "mangohud-amdgpu");
 }
--- a/src/dbus.cpp
+++ b/src/dbus.cpp
@@ -459,7 +459,7 @@ void dbus_manager::start_thread() {
     stop_thread();
     m_quit = false;
     m_thread = std::thread(&dbus_manager::dbus_thread, this);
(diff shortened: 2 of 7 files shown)


smcv:
> Thread names are limited to 15 bytes + '\0', so some of them needed to be abbreviated: `mango-gpufdinfo` just about fits, but there isn't enough space to pluralize `mango-fpsmetric`, and similarly `mango-hwinfoupd` has reached the limits of the available space.

I considered using names like `mangohud-dbus` in shorter cases, but there isn't enough space to do that for the longer names, so some of them would have to be `mango-` anyway, and it seemed maybe better to be consistent.

It would be possible to save one more byte by not using a separator character, perhaps combined with capitalization for clarity - `MangoAmdGpu`, `MangoDBus`, `MangoFpsMetrics`, etc.? - but `MangoHwInfoUpd` probably still looks better than adding one more byte to it.

flightlessmango:
I don't think just mango is specific enough, just setting gpufd, hwinfo, fpsmet should be enough so we can fit mangohud.
I'd think it's more important to know where it's coming from than that the thread is perfectly specified since that can be looked up in the source code.

flightlessmango:
Yeah that looks fine to me

smcv:
> I'd think it's more important to know where it's coming from than that the thread is perfectly specified since that can be looked up in the source code.

Yes, the part after the common prefix only needs to be unique within Mangohud.

Issue #1667, opened 2025-04-29
libMangoHud_shim.so exports extra symbols such as eh_hash_elf, __dlsym, get_proc_address


smcv:
**Describe the bug**
`libMangoHud_shim.so` is a `LD_PRELOAD` module, so every symbol that it exports will "interpose" symbols of the same name in any other library, and be called instead of the other implementation.

It looks as though it is exporting more symbols than it strictly needs to, which risks crashes or other badness.

**List relevant hardware/software information**
- Linux Distribution: anything recent, tested on Arch
- MangoHud version: 0.8.1-1 or git dc309505
- GPU: AMD Vega

**To Reproduce**
`objdump -T -x /path/to/libMangoHud_shim.so`

**Expected behavior**
Only the symbols that MangoHud is intentionally replacing are exported, namely those declared with `CREATE_FWD_VOID` or `CREATE_FWD` and hooked with `ADD_HOOK`:
* `glXSwapBuffers`
* `eglGetDisplay`
* `eglGetPlatformDisplay`
* `eglSwapBuffers`
* `glXSwapBuffersMscOML`
* `glXGetProcAddress`
* `glXGetProcAddressARB`
* `eglGetProcAddress`
plus any that are handled specially:
* `dlsym`
* (others?)

**Actual behavior**

These symbols are exported:

```
DYNAMIC SYMBOL TABLE:
...
0000000000006018 g    DO .bss   0000000000000008  Base        __dlsym
0000000000003aa0 g    DF .text  00000000000000ab  Base        dlsym
0000000000002810 g    DF .text  0000000000000059  Base        eh_hash_elf
0000000000003730 g    DF .text  000000000000008a  Base        real_dlsym
00000000000037c0 g    DF .text  000000000000004d  Base        glXSwapBuffers
00000000000029e0 g    DF .text  0000000000000041  Base        eh_hash_gnu
0000000000002f30 g    DF .text  0000000000000111  Base        eh_set_rel
0000000000003860 g    DF .text  0000000000000065  Base        eglGetPlatformDisplay
0000000000002610 g    DF .text  00000000000001aa  Base        eh_init_obj
0000000000003810 g    DF .text  0000000000000050  Base        eglGetDisplay
0000000000003b50 g    DF .text  0000000000000014  Base        get_proc_address
0000000000003b70 g    DF .text  000000000000008e  Base        eh_iterate_callback
0000000000003a50 g    DF .text  0000000000000050  Base        eglGetProcAddress
[...]

PR #1668, opened 2025-04-30
shim: Avoid exporting unintended symbols

smcv:
* shim: Reuse Mesa's macros instead of open-coding
    
    We'll want macros.h for PUBLIC, and it defines an ARRAY_SIZE that we
    can use.

* shim: Explicitly export forwarding functions
    
    This will allow us to control exports with -fvisibility=hidden.

* shim: Explicitly export our dlsym() reimplementation
    
    This is intended to be hooked in the process that loaded MangoHud.

* shim: Hide all symbols that are not explicitly public
    
    This avoids accidentally interposing symbols like __dlsym() and
    eh_hash_elf().
    
    Resolves: https://github.com/flightlessmango/MangoHud/issues/1667

* real_dlsym: Make local variables static
    
    There's no need for these to be referenced outside this translation unit.

### DIFF
--- a/src/gl/shim.c
+++ b/src/gl/shim.c
@@ -4,6 +4,7 @@
 #include <stdio.h>
 #include <stdint.h>
 #include "gl.h"
+#include "mesa/util/macros.h"
 #include "real_dlsym.h"
 #include <string.h>
 #include <stdbool.h>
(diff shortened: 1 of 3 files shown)


smcv:
Needs testing, I'll try to do that later today.

BananaWorks07:
Any updates on this?

PR #1673, opened 2025-05-05
add qualcomm kgsl support

17314642:
only gpu usage, temperature and core clock is supported

### DIFF
--- a/src/gpu.cpp
+++ b/src/gpu.cpp
@@ -71,18 +71,25 @@ GPUS::GPUS(overlay_params* params) : params(params) {
             }
         }
 
+        std::string msm_driver;
         if (!vendor_id) {
             auto line = read_line("/sys/class/drm/" + node_name + "/device/uevent" );
             if (line.find("DRIVER=msm_dpu") != std::string::npos) {
                 SPDLOG_DEBUG("MSM device found!");
                 vendor_id = 0x5143;
+                msm_driver = "msm";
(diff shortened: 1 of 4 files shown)


src/gpu.cpp
```diff
@@ -76,6 +76,11 @@ GPUS::GPUS(overlay_params* params) : params(params) {
             if (line.find("DRIVER=msm_dpu") != std::string::npos) {
                 SPDLOG_DEBUG("MSM device found!");
                 vendor_id = 0x5143;
+                device_id = 0;
+            } else if (line.find("DRIVER=msm_drm") != std::string::npos) {
+                SPDLOG_DEBUG("MSM kgsl device found!");
+                vendor_id = 0x5143;
+                device_id = 1;
```
flightlessmango, line 83: Don't repurpose `device_id`. You can probably just set `driver` here already?

17314642: fixed?

flightlessmango: yes that's better. But should probably also set `device_id` if we have that information

17314642: panfrost detection has a joke number 1337, because if im correct arm hardware doesnt use the concept of vendor and device id

when i googled clinfo output of some mali gpus, they report 0 for device id

17314642: so that's why i wanted to make driver name the main way to detect gpus, and make vendor and device id secondary attributes for gpus that actually use them (amd and nvidia)

PR #1675, opened 2025-05-08
add qcom cpu temp support via thermal zones

17314642:


flightlessmango:
can you rebase this so the werror test is applied

PR #1677, opened 2025-05-09
Sed Fix

ring-c:
On Linux Mint we have a `/etc/upstream-release` dir, so every launch I got a error:
```
sed: read error on /etc/upstream-release: Is a directory
```
This PR fixes it.

### DIFF
--- a/src/overlay.cpp
+++ b/src/overlay.cpp
@@ -771,7 +771,7 @@ void init_system_info(){
       trim(cpu);
       kernel = exec("uname -r");
       trim(kernel);
-      os = exec("sed -n 's/PRETTY_NAME=\\(.*\\)/\\1/p' /etc/*-release");
+      os = exec("sed -n 's/PRETTY_NAME=\\(.*\\)/\\1/p' /etc/os-release");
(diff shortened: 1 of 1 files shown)


flightlessmango:
The old and new doesn't produce any output for me

ring-c:
Option `-n` suppress output. In normal operation there must be no output.

flightlessmango:
On arch linux, this leads us to /etc/arch-release which is an empty file

ring-c:
I am just fixed an error, not trying to change behavior.

flightlessmango:
What you've proposed destroys the functionality on arch at least.
We need a solution that not only removes the error but also works universally

ring-c:
What is changes for arch if it has empty file anyway?

flightlessmango:
Like I said before `/etc/os-release` is widely adopted, so instead of using `find`, just setting it to sed that file should resolve this for arch, mint and most distros

ring-c:
Commited new change

flightlessmango:
Need to squash this into one commit, otherwise looking good

PR #1678, opened 2025-05-09
Adds font_size_secondary parameter

jbellionjourdan:
Creates a font_size_secondary parameter to allow to chose font size for the "less important" metrics like resolution, present mode, device battery...

Those metrics were displayed with the same small font than the units, which can become very small as the main font size goes down.

If font_size_secondary isn't supplied, it is computed as `0.55 * font_size` which is the same ratio applied to the existing small font.

### DIFF
--- a/README.md
+++ b/README.md
@@ -378,8 +378,9 @@ Parameters that are enabled by default have to be explicitly disabled. These (cu
 | `font_glyph_ranges`                | Specify extra font glyph ranges, comma separated: `korean`, `chinese`, `chinese_simplified`, `japanese`, `cyrillic`, `thai`, `vietnamese`, `latin_ext_a`, `latin_ext_b`. If you experience crashes or text is just squares, reduce font size or glyph ranges |
 | `font_scale=`                      | Set global font scale. Default is `1.0`                                               |
 | `font_scale_media_player`          | Change size of media player text relative to `font_size`                              |
-| `font_size=`                       | Customizable font size. Default is `24`                                              |
-| `font_size_text=`                  | Customizable font size for other text like media metadata. Default is `24`           |
+| `font_size=`                       | Customizable font size. Default is `24`                                               |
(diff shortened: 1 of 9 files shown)


jbellionjourdan:
> what happened with README.md?

My bad. My stupid IDE absolutely wanted to reformat these tables, didn't notice i was committing that.

17314642:
rebase that so that there is no fix readme commit

.gitignore
```diff
@@ -49,3 +49,5 @@ subprojects/cmocka/
 **/GPATH
 **/GRTAGS
 **/GTAGS
+
+.idea/
```
flightlessmango, line 53: This seems irrelevant

build.sh
```diff
@@ -54,7 +54,7 @@ dependencies() {
         for i in $DISTRO; do
         echo "# Checking dependencies for \"$i\""
         case $i in
-            *arch*|*manjaro*|*artix*|*SteamOS*)
+            *arch*|*manjaro*|*artix*|*SteamOS*|*garuda*)
```
flightlessmango, line 57: This should be a different commit or even a different PR

jbellionjourdan: Fine by me

src/hud_elements.h
```diff
@@ -117,6 +117,7 @@ class HudElements{
         static void winesync();
         static void present_mode();
         static void network();
+
```
flightlessmango, line 120: don't add random new lines

jbellionjourdan:
Hi guys, anything missing here for you ?

flightlessmango:
The "merge" commit should not be part of the PR

jbellionjourdan:
Done. Branch rebased properly.

flightlessmango:
I merged a bunch of stuff at the same with this PR, something probably caused a conflict so I reverted it.
Can you re-submit this?

PR #1680, opened 2025-05-11
overlay_params: add additional known handhelds to exclusion list

matte-schwartz:
Cleans up case 4 (full) on known handhelds like the Legion Go S, Legion Go, and ROG Ally.

| Before | After |
|--------|--------|
|go-s-full-unpatched-20250511123452|go-s-full-patched-20250511123620|
|go-full-unpatched-20250511123946|go-full-patched-20250511124011|

I removed `gpu_power_limit` since it didn't seem to be working on Rembrandt or Phoenix, but that could possibly be a bug. Not sure.

### DIFF
--- a/src/overlay_params.cpp
+++ b/src/overlay_params.cpp
@@ -1210,14 +1210,18 @@ void presets(int preset, struct overlay_params *params, bool inherit) {
          if (!gpus)
             gpus = std::make_unique<GPUS>(params);
          
-         // Disable some options if steamdeck
+         // Disable some options if steamdeck / other known handhelds
          for (auto gpu : gpus->available_gpus) {
-            if (gpu->device_id == 0x1435 || gpu->device_id == 0x163f){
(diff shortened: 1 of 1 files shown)


src/overlay_params.cpp
```diff
@@ -1219,6 +1219,15 @@ void presets(int preset, struct overlay_params *params, bool inherit) {
                add_to_options(params, "gpu_mem_temp", "0");
                add_to_options(params, "gpu_efficiency", "0");
             }
+            // Rembrandt and Phoenix APUs (Z1, Z1E, Z2 Go)
+            else if (gpu->device_id == 0x1681 || gpu->device_id == 0x15bf){
```
flightlessmango, line 1223: I'd prefer having the joint options under the same if and then a separate statement for those that are unit specific

matte-schwartz: makes sense, fixed

Issue #1243, opened 2024-02-20
Thinks that I'm constantly thermal throttling / power limited


LivacoNew:
**Describe the bug**
Recently MangoHUD has been incorrectly displaying that I'm constantly thermal throttling. On top of this, it at times displays that I'm power limited even though from what I can tell my CPU/GPU are completely fine. 

**List relevant hardware/software information**
- Arch Linux
- MangoHud v0.7.1
- Mesa 24.0.1-arch1.1
- AMD Ryzen 7 7800X3D / Sapphire Radeon RX 7800XT Nitro+ 

**To Reproduce**
1. Play anything with MangoHUD enabled.

**Expected behavior**
MangoHUD displays if I'm throttled or power limited correctly.

**Screenshots**
VKCube is not enough to throttle this system in the slightest. Btop on the bottom left confirms my CPU is at a perfectly fine temp.
![image](https://github.com/flightlessmango/MangoHud/assets/51921257/d1db0e18-771f-4cff-9274-9b58a1e39458)

**Additional context**
N/A

flightlessmango:
Unrelated to this issue, it looks like you have two instances of mangohud installed.
It appears to be rendering twice

LivacoNew:
I think I just took the screenshot right as it was updating - I only have the one from the official arch repo and that's it

flightlessmango:
We are correctly interpreting the information the driver gives us.
This had to be corrected before upstream https://github.com/flightlessmango/MangoHud/issues/748#issuecomment-1653315797
Since this GPU is not in NAVI22/23/24 batch, it's likely it needs to be fixed upstream as well

GStar718:
does this still work?

flightlessmango:
> does this still work?

I don't know what you mean exactly

LivacoNew:
> does this still work?

I'm still getting this issue if that's what you meant.

LivacoNew:
^ Seems the guy who reported the issue mentioned above has reported this to their [Gitlab](https://gitlab.freedesktop.org/drm/amd/-/issues/3251) with other people also having the issue confirming it's a amdgpu's problem.

Umio-Yasuno:
This is clearly a firmware and the amdgpu driver issue, but a workaround may need to be implemented in user space.  

https://gitlab.freedesktop.org/drm/amd/-/issues/3251#note_2904104

LivacoNew:
[...]

PR #1690, opened 2025-05-18
amdgpu: implement the invalid TEMP_HOTSPOT throttling workaround for RDNA 3 dGPU & some fixes

Umio-Yasuno:
RDNA 3 dGPU will show invalid TEMP_HOTSPOT throttling, but as it is unlikely to be fixed on the driver side at this time, we will implement a workaround.  

<https://gitlab.freedesktop.org/drm/amd/-/issues/3251#note_2904104>

This PR also includes fixes for minor issues found during the work.  

#1243

### DIFF
--- a/src/amdgpu.cpp
+++ b/src/amdgpu.cpp
@@ -78,8 +78,15 @@ void AMDGPU::get_instant_metrics(struct amdgpu_common_metrics *metrics) {
 		metrics->current_uclk_mhz = amdgpu_metrics->current_uclk;
 
 		metrics->gpu_temp_c = amdgpu_metrics->temperature_edge;
-		indep_throttle_status = amdgpu_metrics->indep_throttle_status;
 		metrics->fan_speed = amdgpu_metrics->current_fan_speed;
+
+		indep_throttle_status = amdgpu_metrics->indep_throttle_status;
+		bool temp_hotspot_flag = ((indep_throttle_status >> TEMP_HOTSPOT_BIT) & 1) == 1;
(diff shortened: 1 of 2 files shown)


src/amdgpu.cpp
```diff
@@ -78,8 +78,15 @@ void AMDGPU::get_instant_metrics(struct amdgpu_common_metrics *metrics) {
...
 
 		metrics->gpu_temp_c = amdgpu_metrics->temperature_edge;
-		indep_throttle_status = amdgpu_metrics->indep_throttle_status;
 		metrics->fan_speed = amdgpu_metrics->current_fan_speed;
+
+		indep_throttle_status = amdgpu_metrics->indep_throttle_status;
+		bool temp_hotspot_flag = ((indep_throttle_status >> TEMP_HOTSPOT_BIT) & 1) == 1;
+		// ThrottlingPercentage for TEMP_HOTSPOT on RDNA 3 dGPU is almost always greater then or equal to 1.
+		// So the AMDGPU driver set the TEMP_HOTSPOT bit even in idle state.
+		if (temp_hotspot_flag && amdgpu_metrics->temperature_hotspot < 90) {
```
flightlessmango, line 87: We don't *know* if it's actually throttling, so should we say it's throttling just because it's above 90c?
Junction can easily reach about 100c without throttling, I'm pretty sure.
Maybe just ignore this case until it's fixed?

flightlessmango: Yeah I think so, just clear it until we can use it properly

Umio-Yasuno: Done. https://github.com/flightlessmango/MangoHud/pull/1690/commits/5d2ec01b4776dec783dc08974d51bb3bd2bdc516

PR #1697, opened 2025-05-27
GPUS: Ensure gpu is not null

Etaash-mathamsetty:
fixes constant exception spam in logs when using winewayland, which makes the game refuse to boot.

### DIFF
--- a/src/gpu.h
+++ b/src/gpu.h
@@ -190,7 +190,7 @@ class GPUS {
 
             if (!params->pci_dev.empty()) {
                 for (auto &gpu : available_gpus) {
-                    if (gpu->pci_dev == params->pci_dev) {
+                    if (gpu && gpu->pci_dev == params->pci_dev) {
                         vec.push_back(gpu);
                         return vec;
                     }


src/gpu.h
```diff
@@ -190,7 +190,7 @@ class GPUS {
 
             if (!params->pci_dev.empty()) {
                 for (auto &gpu : available_gpus) {
-                    if (gpu->pci_dev == params->pci_dev) {
+                    if (gpu && gpu->pci_dev == params->pci_dev) {
```
flightlessmango, line 193: Why is there a NULL gpu added to `available_gpus`?
Would be great to find the root cause

Etaash-mathamsetty: It turns out it got fixed between commit `d297a57fc2ca2e5be1bb63208154701a277ab06c` and the latest master, and then when I tried to test it using my fix it was already fixed since I built from master instead of `d297a57fc2ca2e5be1bb63208154701a277ab06c`. I assume it was probably caused by some weird race condition or something

PR #1701, opened 2025-05-30
shim: Capture dlerror from at start of dlsym to restore later

FoxieFlakey:
If `real_dlsym(handle, name)` succeded but earlier dlsym on eglStreamPostD3DTextureANGLE failed it
will create non conforming behaviour due dlsym won't clear the error message.

An example of real issue is Minecraft lost audio when MangoHud is used due OpenAL's perfectly
conformance usage like below

```cpp
    dlerror();
    auto *sym = dlsym(handle, name);
    // Obviously OpenAL only loads audio libraries so eglStreamPostD3DTextureANGLE obviously fails
    // and erronously sets error message even if 'name' exists
    if(auto *err = dlerror(); err != nullptr)
        sym = nullptr;
    return sym;
```
Quoted from https://github.com/kcat/openal-soft/blob/7f56dcdfbce02bea96894c688d48be574a55a41d/common/dynload.cpp#L38-L45
and added comment

This is perfectly valid as first call clear the error message, then perform dlsym. If it succeded
error message won't be set but not be cleared either. The standard only describes that error will be
set if there is any error not if there successes.

Therefore MangoHud indirectly/directly through dlsym breaks OpenAL and other dlsym users which follows
similar pattern assumption.

~~Side notes:
There is still conformance issue when there was an error before hook is called but may be
overwritten by dlsym for eglStreamPostD3DTextureANGLE but it is ignored for now due then
need a way to set dlerror message to restore the erorr message but I haven't found example
that new problem causing issues.~~ (fixed this)

dlerror reference: https://pubs.opengroup.org/onlinepubs/009604299/functions/dlerror.html

### DIFF
--- a/src/gl/shim.c
+++ b/src/gl/shim.c
@@ -185,16 +185,78 @@ static struct func_ptr hooks[] = {
 #define ARRAY_SIZE(arr) sizeof(arr)/sizeof(arr[0])
 #endif
 
+#include <stddef.h>
+
+// Glibc has nonconformance behaviour
+// which is Glibc clears dlerror if dlsym succeded instead
+// leaving it as it is as required by standard
+// Reference: https://pubs.opengroup.org/onlinepubs/009604299/functions/dlsym.html
+//
+// Lets disable dlerror override to avoid risking programs
+// which already relying on Glibc's behaviour so to keep
+// glibc's behaviour
+// while on conformance system such as Musl it fixes bugs
+#ifdef __GLIBC__
+# define USE_DLERROR_OVERRIDE 0
+#else
+# define USE_DLERROR_OVERRIDE 1
+#endif
+
+static _Thread_local char *error_override = NULL;
+static _Thread_local bool activate_override = false;
+
+char *dlerror(void)
+{
+    if (!USE_DLERROR_OVERRIDE)
+       return real_dlerror();
+    
+    if (activate_override)
+    {
+        // Deactivate the override for next call to allow
+        // real_dlerror to be used
+        activate_override = false;
+        return error_override;
+    } else if (error_override) {
+        // Free resources that was allocated for error override
+        // because the override is now deactivated
+        free(error_override);
+        error_override = NULL;
+    }
+
+    return real_dlerror();
+}
+
+static void save_and_consume_real_dlerror()
+{
+    if (!USE_DLERROR_OVERRIDE)
+       return;
+    
+    char* recent_error = real_dlerror();
+    if (!recent_error)
+    {
+        error_override = recent_error;
+        return;
+    }
+    
+    size_t error_length = strlen(recent_error);
+    error_override = malloc(error_length + 1);
+    memcpy(error_override, recent_error, error_length + 1);
+}
+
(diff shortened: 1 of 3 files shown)


FoxieFlakey:
I made a small test program to test `dlsym` behaviour for comparison. This PR allows MangoHud's `dlsym` to behave correctly according to standard on Musl platform and leaves don't touch glibc behaviour at all

# Expected results:
## Glibc:
No MangoHud: Test 1 failed (Glibc's nonconforming behaviour[2])
MangoHud: Test 1 failed. Test 2 failed[1]
With this patch: Test 1 failed (Preserving libc behaviour)

## Musl (and also standard behaviour)
No MangoHud: All test passes
MangoHud: Test 1 failed. Test 2 failed[1]
With this patch: All test passes (Preserving libc behaviour)

# Notes:
1. Test 2 must be manually checked (due nature of error message might be changing), the error must be referring to 'this_function obviously do not exist UwU' symbol not "eglStreamPostD3DTextureANGLE". MangoHud overwritten the error message with different error.
2. https://pubs.opengroup.org/onlinepubs/009604299/functions/dlerror.html states last error occured so it should not be erased except by `dlerror` but glibc clears it and creates non conforming behaviour which the PR must preserve as it could break programs relying on that behaviour

The test program compile it with `cc -o <output> <the source file>.c`
```c
#include <dlfcn.h>
#include <stdio.h>
#include <stddef.h>

static void test1() {
  printf("[Test 1] dlsym on existing symbol with initially cleared error\n");
  // Clear errs
  dlerror();
  dlsym(NULL, "dlerror");
  char* err = dlerror();
  if (err == NULL)
    printf("[Test 1] Success\n");
  else
    printf("[Test 1] Failed, dlerror must return NULL: Error was: '%s'\n", err);
}

static void test2() {
  printf("[Test 2] dlsym on existing symbol with preexisting error\n");
  // Clear errs
  dlerror();
  dlsym(NULL, "this_function obviously do not exist UwU");
  dlsym(NULL, "dlerror");
  char* err = dlerror();
  if (err != NULL)
    printf("[Test 2] Undetermined please check: Error was: '%s'\n", err);
  else
    printf("[Test 2] Failed, dlerror must return non NULL\n");
}

static void test3() {
  printf("[Test 3] dlsym on unknown symbol with initially cleared error\n");
  // Clear errs
  dlerror();
  dlsym(NULL, "this_sym_dont_exist");
  char* err = dlerror();
  if (err == NULL)
    printf("[Test 3] Failed, dlerror must return NULL: Error was: '%s'\n", err);
  else
    printf("[Test 3] Success\n");
}

static void test4() {
  printf("[Test 4] dlsym on unknown symbol with preexisting error\n");
  // Clear errs
  dlerror();
  dlsym(NULL, "this_function obviously do not exist UwU");
  dlsym(NULL, "this_sym_dont_exist");
  char* err = dlerror();
  if (err != NULL)
    printf("[Test 4] Success: Error was: '%s'\n", err);
  else
    printf("[Test 4] Failed, dlerror must return non NULL\n");
}

int main() {
  test1();
  test2();
  test3();
  test4();
}
```

PR #1706, opened 2025-06-02
ftrace: add support

zdobersek:
Add the ability to display information gathered from ftrace events using the 'ftrace' option. Currently this information can be displayed in three ways:
- histograms, displaying the number of occurrences of a given event. Example: 'histogram/drm_vblank_event'
- line graphs, plotting value of a given event parameter. Example: 'linegraph/devfreq_monitor/freq' will plot the value of the renderpasses parameter as reported through the devfreq_monitor event.
- labels, simply displaying the value of a given tracepoint parameter. Example: 'label/devfreq_frequency/dev_name'

More than one point of information can be displayed by combining them in the 'ftrace' option:
  ftrace=histogram/start_cmd_buffer+label/start_cmd_buffer/IR3debugFlags

Usage of this feature has certain conditions and limitations:
- tracefs has to be mounted so that the user running MangoHud has the necessary permissions to access ftrace interface,
- ftrace still has to be configured manually, e.g. the observed events have to be enabled, can be filtered etc.,
- only one instance of MangoHud can parse and display these events because only one consumer can read from tracefs' trace_pipe,
- current parsing logic depends on fields to be formatted in the `key=value` fashion, which isn't standardized but is most common formatting choice.


zdobersek:
Added `HAVE_FTRACE` guards for (most of) the introduced code. The macro is defined on Linux. I can switch to simply using `__linux__` but it's somewhat generic IMO.

PR #1725, opened 2025-06-21
Fix GTK blacklist

Etaash-mathamsetty:
This should fix globally enabled mangohud with GTK 4 applications. It would be nice if this could get tested as I currently have no way of testing this tho. 

Source for engine name: https://github.com/GNOME/gtk/blob/13d6c41fc56e2cfa754a85eb0116efd3db22a284/gdk/gdkvulkancontext.c#L1817C37-L1817C48

### DIFF
--- a/src/blacklist.cpp
+++ b/src/blacklist.cpp
@@ -10,6 +10,7 @@
 
 namespace fs = ghc::filesystem;
 std::string global_proc_name;
+std::string global_engine_name;
 
 static std::string get_proc_name() {
    // Note: It is possible to use GNU program_invocation_short_name.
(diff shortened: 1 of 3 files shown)


flightlessmango:
I want to remember when I checked, it lists engine as vulkan

BananaWorks07:
> I want to remember when I checked, it lists engine as vulkan

That's because vulkan is the engines that are everything other than dxvk, vkd3d, etc

flightlessmango:
Right so this wouldn't work then?

BananaWorks07:
> Right so this wouldn't work then?

I'm comparing the original engine name string instead of the enum's value so it should still work

Issue #1649, opened 2025-04-08
gpu_fan option to show percent for Nvidia cards


luisalvarado:
Currently testing the latest compiled version of MangoHud

But for Nvidia cards like the 4070, 4090 or 5090, it shows the RPM by default. Was curious to know if there is an option to show the Percent version of it for Nvidia cards. Currently it shows as

![Image](https://github.com/user-attachments/assets/c4b89222-6e9b-491c-8106-131540658fbc)

luisalvarado:
Hi, it shows in percent.

![Image](https://github.com/user-attachments/assets/4f63856b-b15f-4de8-86a7-ccbf1364c7c8)

![Image](https://github.com/user-attachments/assets/fce6eaa7-3d64-4599-ac4d-ef69b848200b)

![Image](https://github.com/user-attachments/assets/72ee25f4-d51d-49b7-9b61-f232deb794dc)

I am also able to change the fan speed using a percent value as seen below if that helps (I use it on a script to set the fan speed depending on the temperature)

nvidia-settings -a "[gpu:0]/GPUFanControlState=1"
nvidia-settings -a "[fan:0]/GPUTargetFanSpeed=50"
nvidia-settings -a "[fan:1]/GPUTargetFanSpeed=50"

17314642:
Can you compile mangohud with `-Dwith_xnvctrl=disabled` so mangohud will use only nvml for fan speed

Does it show in rpm now?

luisalvarado:
I did it like this

```./build.sh build -Dwith_xnvctrl=disabled```

but still shows with RPMs, maybe I am supposed to add the parameter somewhere else?

Also just noticed this warnings for future reference:

Found ninja-1.12.1 at /usr/bin/ninja
**WARNING: Running the setup command as `meson [options]` instead of `meson setup [options]` is ambiguous and deprecated.**                                                                                                                                                                                                                                       
The Meson build system
Version: 1.7.0

Found ninja-1.12.1 at /usr/bin/ninja
**WARNING: Running the setup command as `meson [options]` instead of `meson setup [options]` is ambiguous and deprecated.**                                                                                                                                                                                                                                       
ninja: Entering directory `build/meson64'

Compiler for C supports link arguments -Wl,-z,relro: YES 
Compiler for C supports link arguments -Wl,--exclude-libs,ALL: YES 
Compiler for C supports link arguments -lGL: NO 
Compiler for C supports link arguments -static-libstdc++: YES 
[...]

PR #1734, opened 2025-06-28
remove xnvctrl support

17314642:
Like discussed in #1649 (https://github.com/flightlessmango/MangoHud/issues/1649#issuecomment-2787022043)


flightlessmango:
xnvctrl is still needed for old nvidia cards

17314642:
how old are we talking? i thought that all gpus starting from gtx 600's are using proprietary nvidia drivers and have nvml library support, no?

flightlessmango:
I thought we were still getting metrics from nvctrl, but looks like it is just fan speed now.. so we can probably nuke it

luisalvarado:
I can check a 1660 I have. What should I check for if that is useful?

17314642:
> I can check a 1660 I have. What should I check for if that is useful?

1660 is too new, we need gtx 600's

MrMeshok:
Do you still need testing with old gpu? I have an old notebook with GTX 620m, I think it's only supports driver version 470 which is not working on newer kernels, but I remember that there is PPAs that work with them

luisalvarado:
The oldest I have is my baby 1080 but that was from 2014 (11 years ago). And it was more expensive than the current low end ones right now. If there was some voting for this then I would vote to note support my arcane, spider wbe crawling card and down.

PR #1744, opened 2025-07-04
Reallow X11 keybind fallback

### LINKED ISSUE #1743: Mangohud ignoring numpad keybinds post #1705
**Describe the bug**
Mangohud no longer recognizes numpad KP_* keybinds, possibly others as well, in certain applications. The hud will still open when bound to a 'normal' keys such as 'p'.

- vkcube: works
- steam native game (cs2): fails to open
- steam proton game (sekiro): fails to open

After reverting  #1705 the problem disappears.

- Gentoo
- Mangohud Git
- 7900xtx
- Wooting keyboard, which does not have a numpad, with binds to send numpad keys. I don't have a keyboard with a numpad. Reproducing this with one would be useful here to rule out any weird edge case.

**To Reproduce**
1. Change toggle_hud in the config to 'toggle_hud=KP_6'.
2. Attempt to open earlier applications via KP_6.

nJ3ahxac:
A refactor to the wayland input system in #1705 inverted some logic and didn't allow a fallthrough when keys were not detected by wayland but could still be detected by X11.

Fixes #1743

### DIFF
--- a/src/keybinds.h
+++ b/src/keybinds.h
@@ -25,15 +25,11 @@ static inline bool keys_are_pressed(const std::vector<KeySym>& keys)
     {
         update_wl_queue();
 
-        if (wl_pressed_keys.size() != keys.size())
-            return false;
-
-        for (KeySym ks : keys) {
-            if (wl_pressed_keys.count(ks) == 0)
-                return false;
+        if (std::all_of(std::begin(keys), std::end(keys), [&](const auto& key) {
+                return wl_pressed_keys.count(key) != 0;
+        })) {
+                return true;
         }
-
-        return true;
     }
     #endif
 


flightlessmango:
@Etaash-mathamsetty

BananaWorks07:
I would rather have this fixed in the Wayland input code. Might be a keyboard layout issue on your system potentially? I'll have to test this.

Also the Wayland code should not be running on xwayland applications, you aren't using xwayland right now correct?

nJ3ahxac:
I am on KDE and afaik it uses wayland when it can and xwayland when necessary, really no expert. When I build mangohud both `HAVE_X11` and `HAVE_WAYLAND` are defined however.

I put some logging in `keys_are_pressed`. This is built with my PR, so it tries wayland first and then tries X11. I just printed the keysym, so it's a bit ugly. The log looks like this when I don't press anything:
```
keys_are_pressed keys: 65462
    wayland keys:
        wayland keys not pressed! trying X11
    x11 keys:
        X11 keys not pressed! returning false
```
When I press the macro it prints this:
```
keys_are_pressed keys: 65462
    wayland keys: 65432
        wayland keys not pressed! trying X11
    x11 keys: 65462
        X11 keys were pressed
```
Here's what `xev` says when I press and release the macro:
```
KeyPress event, serial 39, synthetic NO, window 0x3e00001,
    root 0x36e, subw 0x0, time 4280609, (107,32), root:(2898,482),
    state 0x0, keycode 85 (keysym 0xff98, KP_Right), same_screen YES,
    XLookupString gives 0 bytes:
    XmbLookupString gives 0 bytes:
    XFilterEvent returns: False

KeyRelease event, serial 39, synthetic NO, window 0x3e00001,
    root 0x36e, subw 0x0, time 4280698, (107,32), root:(2898,482),
    state 0x0, keycode 85 (keysym 0xff98, KP_Right), same_screen YES,
    XLookupString gives 0 bytes:
    XFilterEvent returns: False
```
Now with wev:
```
                      sym: KP_Right     (65432), utf8: ''
[        16:     wl_keyboard] key: serial: 19830; time: 4368045; key: 85; state: 0 (released)
                      sym: KP_Right     (65432), utf8: ''
[        16:     wl_keyboard] key: serial: 19833; time: 4368430; key: 64; state: 1 (pressed)
```

So there's some strange inconsistency with games launched with steam where X11 and wayland disagree on the state of numlock, even though wev and xev agree. It seems like steam doesn't respect the numlock state of the environment and just sets it to off, which produces `KP_Right` instead of `KP_6`. 

The fix here was to change all of my key bindings to use their alternative states (eg, instead of 'KP_6' I used 'KP_Right'), and to tell KDE to start with numlock off with: "System Settings -> Input & Output -> Keyboard -> NumLock on Startup = Turn Off".  Now everything functions as expected without the PR.

I think this PR should be closed as from what I can see, the bug lies with Steam not respecting the state of numlock in the environment. Maybe this issue could just be documented for now as a quirk.

PR #1775, opened 2025-08-09
Fix ntsync detection

Etaash-mathamsetty:
Fixes ntsync detection on cachyos proton

### DIFF
--- a/src/winesync.h
+++ b/src/winesync.h
@@ -55,7 +55,7 @@ class WineSync {
                     method = syncMethods::NTSYNC;
                     else if (sym.find("fsync") != std::string::npos)
                         method = syncMethods::FSYNC;
-                    else if (sym.find("ntsync") != std::string::npos)
+                    else if (sym.find("/dev/ntsync") != std::string::npos)
                         method = syncMethods::NTSYNC;
                     else if (sym.find("esync") != std::string::npos)
                         method = syncMethods::ESYNC;


flightlessmango:
I don't see why this would change anything, if `/dev/ntsync` is in the string, so is `ntsync`

BananaWorks07:
> I don't see why this would change anything, if `/dev/ntsync` is in the string, so is `ntsync`

Looking through mapped fd would also look through stderr and stdout. Cachyos proton prints `ntsync explicitly disabled` in stderr

17314642:
> > I don't see why this would change anything, if `/dev/ntsync` is in the string, so is `ntsync`
> 
> Looking through mapped fd would also look through stderr and stdout. Cachyos proton prints `ntsync explicitly disabled` in stderr

sync detection code doesnt read data inside fds, it only resolves symlinks and searches for key words in file path

17314642:
run the program below on the problematic game and send the output

```c++
#include <iostream>
#include <filesystem>
#include <limits.h>
#include <unistd.h>
#include <array>

std::string read_symlink(const char* link) {
    char result[PATH_MAX] {};
    ssize_t count = readlink(link, result, PATH_MAX);
    return std::string(result, (count > 0) ? count : 0);
}

void check_dir(const std::string& dir) {
    for (auto& p : std::filesystem::directory_iterator(dir)) {
        const std::string filepath = p.path().string();
        std::string sym = read_symlink(filepath.c_str());

        for (const std::string s : {"winesync", "fsync", "ntsync", "esync"}) {
            if (sym.find(s) == std::string::npos)
                continue;

            std::cout << "symlink: \"" << filepath << "\" resolved: \"" << sym << "\" type: " << s << std::endl;
        }
    }
}

int main(const int argc, const char* argv[]) {
    if (argc < 2) {
        std::cout << "specify pid\n";
        return -1;
    }

    std::string pid_dir = "/proc/" + std::string(argv[1]);

    if (!std::filesystem::exists(pid_dir)) {
        std::cout << "process doesn't exist\n";
        return -1;
    }

    std::array<std::string, 2> paths;
    paths[0] = pid_dir + "/map_files";
    paths[1] = pid_dir + "/fd";

    try {
        for (const std::string& s : paths)
            check_dir(s);
    } catch(const std::exception& ex) {
        std::cout << "caught exception: " << ex.what() << std::endl;
    }

    return 0;
}
```

BananaWorks07:
> run the program below on the problematic game and send the output
> 
> ```c++
> #include <iostream>
> #include <filesystem>
> #include <limits.h>
> #include <unistd.h>
> #include <array>
> 
> std::string read_symlink(const char* link) {
>     char result[PATH_MAX] {};
>     ssize_t count = readlink(link, result, PATH_MAX);
>     return std::string(result, (count > 0) ? count : 0);
> }
> 
> void check_dir(const std::string& dir) {
>     for (auto& p : std::filesystem::directory_iterator(dir)) {
>         const std::string filepath = p.path().string();
>         std::string sym = read_symlink(filepath.c_str());
> 
>         for (const std::string s : {"winesync", "fsync", "ntsync", "esync"}) {
>             if (sym.find(s) == std::string::npos)
>                 continue;
> 
>             std::cout << "symlink: \"" << filepath << "\" resolved: \"" << sym << "\" type: " << s << std::endl;
>         }
>     }
> }
> 
> int main(const int argc, const char* argv[]) {
>     if (argc < 2) {
>         std::cout << "specify pid\n";
>         return -1;
>     }
> 
>     std::string pid_dir = "/proc/" + std::string(argv[1]);
> 
>     if (!std::filesystem::exists(pid_dir)) {
>         std::cout << "process doesn't exist\n";
>         return -1;
>     }
> 
>     std::array<std::string, 2> paths;
>     paths[0] = pid_dir + "/map_files";
>     paths[1] = pid_dir + "/fd";
> 
>     try {
>         for (const std::string& s : paths)
>             check_dir(s);
>     } catch(const std::exception& ex) {
>         std::cout << "caught exception: " << ex.what() << std::endl;
>     }
> 
>     return 0;
> }
> ```

I'll check proton cachyos source code then, might be something silly there

PR #1777, opened 2025-08-12
Use cpu instead of cpu_family for mangohud.json.in

MrMeshok:
In Debian we got this warning

> There are issues with the [multiarch](https://wiki.debian.org/MultiArch) metadata for this package.
> [mangohud conflicts on /usr/share/vulkan/implicit_layer.d/MangoHud.arm.json on armel <-> armhf](https://wiki.debian.org/MultiArch/Hints#file-conflict)

If you try to install both armel and armhf packages you will get this error

> trying to overwrite shared '/usr/share/vulkan/implicit_layer.d/MangoHud.arm.json', which is different from other instances of package mangohud:armhf

After this change this will work.
To test this change I compiled for amd64, i386, armel, armhf and got this results
| arch | cpu | cpu_family |
| ------ | ---- | -------------- |
| amd64 | x86_64 | x86_64 |
| i386 | i386 | x86 |
| armel | armel | arm |
| armhf | armhf | arm |

### DIFF
--- a/src/meson.build
+++ b/src/meson.build
@@ -29,7 +29,7 @@ conf_data = configuration_data()
 
 conf_data.set('ld_libdir_mangohud_abs', libdir_mangohud)
 conf_data.set('ld_libdir_mangohud', ld_libdir_mangohud)
-conf_data.set('cpu_family', host_machine.cpu_family())
+conf_data.set('cpu_family', host_machine.cpu())
 conf_data.set('version', describe_ver)
 
 overlay_shaders = [
@@ -327,7 +327,7 @@ if is_unixy
 endif
 
 configure_file(input : 'mangohud.json.in',
-  output : '@0@.@1@.json'.format(meson.project_name(), host_machine.cpu_family()),
+  output : '@0@.@1@.json'.format(meson.project_name(), host_machine.cpu()),
   configuration : conf_data,
   install : true,
   install_dir : join_paths(get_option('datadir'), 'vulkan', 'implicit_layer.d'),


MrMeshok:
Looking at i386 results now I realized, that install/uninstall scripts should be adjusted. 
I can do this, but maybe this should be Debian only patch 🤔

flightlessmango:
I was racking my brain trying to think of how this could possibly mess with downstream, but I think it should be fine?
I'm curious how other projects tend to handle this

MrMeshok:
https://github.com/flightlessmango/MangoHud/blob/aa4333fbaf79f3c5725f26de4b2c3bde61e8834f/bin/mangohud-setup.sh#L35
https://github.com/flightlessmango/MangoHud/blob/aa4333fbaf79f3c5725f26de4b2c3bde61e8834f/bin/mangohud-setup.sh#L72

This lines would need to be adjusted because now this file would be named "MangoHud.i386.json" and old x86 name should be kept in uninstall so new install/uninstall script can uninstall older version.

I will look into how other distros handle this, but they probably don't have this problem because not many build for armel and armhf

MrMeshok:
I found very few distros that supports armel and didn't based on Debian and they all have the same problem.
But I thought about it more and rewritten this PR to only affect arm 32 bit architectures, so this change will have very small footprint and didn't need any changes for install/uninstall script

MrMeshok:
Also found a similar issue for mesa
https://gitlab.freedesktop.org/mesa/mesa/-/issues/13745

MrMeshok:
Debian dropped armel, so this doesn't really matter anymore

PR #1787, opened 2025-08-25
Manghud pathor drm driver enable

pdapandapda:
As linux 6.10+ uses "Panthor" drm driver with panfrost, the profiling and driver name should be add to mangohud.
remember to  "echo <N> > /sys/bus/platform/drivers/panthor/[a-f0-9]*.gpu/profiling"


pdapandapda:
Works with rk3588 kernel 6.16.1 panvk now

17314642:
why did you remove relevant comments?

pdapandapda:
> why did you remove relevant comments?

PR #1811, opened 2025-10-04
Add a parameter for font_size_small

paypur:


fxzzi:
thanks, just tested and works great on my end.

PR #1751, opened 2025-07-11
independent fps_superscript and engine_names hiding

JustCryen:
Closes https://github.com/flightlessmango/MangoHud/issues/1045

Implemented two new parameters: `hide_fps_superscript` and `hide_engine_names`. 
Engine name and fps superscript appearance are no longer dependent on `hud_compact` and `horizontal` parameters.

This allows for these previously impossible configurations:
- Vertical non-compact with no engine name and no superscript
Screenshot From 2025-07-11 04-06-39  

- Horizontal compact with FPS superscript
Screenshot From 2025-07-11 04-08-09

- Horizontal compact short name with FPS superscript
Screenshot From 2025-07-11 04-07-42  

Obviously all previous configurations can also be achieved by setting these parameters accordingly.

PR #1821, opened 2025-10-13
README & config file update for pr #1751

JustCryen:
As discussed in https://github.com/flightlessmango/MangoHud/pull/1751  
I'm open to suggestions about description phrasing in the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -411,6 +411,8 @@ Parameters that are enabled by default have to be explicitly disabled. These (cu
 | `gpu_list`                         | List GPUs to display `gpu_list=0,1`                                                   |
 | `gpu_efficiency`                   | Display GPU efficiency in frames per joule                                            |
 | `gpu_power_limit`                  | Display GPU power limit                                                               |
+| `hide_engine_names`                | Hides the engine name and replaces it with the colored FPS text                       |
+| `hide_fps_superscript`             | Hides the smaller FPS superscript text to the right of the FPS count                  |
 | `hide_fsr_sharpness`               | Hides the sharpness info for the `fsr` option (only available in gamescope)           |
 | `histogram`                        | Change FPS graph to histogram                                                         |
 | `horizontal`                       | Display Mangohud in a horizontal position                                             |
--- a/data/MangoHud.conf
+++ b/data/MangoHud.conf
@@ -144,6 +144,7 @@ fps
 # fps_value=30,60
 # fps_color=B22222,FDFD09,39F900
 # fps_text=""
+# hide_fps_superscript
 frametime
 # frame_count
 ## fps_metrics takes a list of decimal values or the value avg
@@ -158,6 +159,7 @@ throttling_status
 ### Display miscellaneous information
 # engine_version
 # engine_short_names
+# hide_engine_names
 # gpu_name
 # vulkan_driver
 # wine


JustCryen:
@ChrisLane @flightlessmango  
It's been a while, I'm not sure if you have seen this PR.  
It is linked to the previous one but maybe you haven't got a notification for it.

ChrisLane (approved):
I hadn't seen this but the changes look sensible.

A couple of minor points to consider for potential improvements:
1. FPS superscript still appears for FPS stats in the `fps_metrics` component.
2. `hide_engine_names` does not override `engine_short_names` so if the latter appears in config, the former will not take effect.

JustCryen:
I have noticed the second behavior but I wasn't sure how this should be handled.  
This being said, I will correct this shortly.

flightlessmango:
> hide_engine_names does not override engine_short_names so if the latter appears in config, the former will not take effect.

This has been reworked a bit in e96a0bf5afbc8f01b648a739bfd34379979c23cc
Might have accidentally fixed it?

ChrisLane:
> > hide_engine_names does not override engine_short_names so if the latter appears in config, the former will not take effect.
> 
> This has been reworked a bit in [e96a0bf](https://github.com/flightlessmango/MangoHud/commit/e96a0bf5afbc8f01b648a739bfd34379979c23cc) Might have accidentally fixed it?

Still reproducible as of mangohud-git @ 5399bf1a184ce41ff55125e4a24b978392025d24

JustCryen:
I wanted to take a look at it again and I noticed new inconsistent behavior.  

Currently it seems like I can't get the full engine name no matter what I do, it seems like full engine name is currently locked at hidden engine name state with hidden superscript.  

I can get the intended behavior only when `engine_short_names` is available in the config, but having both `engine_short_names` and `hide_engine_names` results in the same hidden state (no fps superscript here, even though it should be visible)  

I will take a look at it again but I'm sure it wasn't like this last time I implemented these options.

JustCryen:
I think I have found the culprit.  

@flightlessmango 
```
if (en[OVERLAY_PARAM_ENABLED_horizontal] && !en[OVERLAY_PARAM_ENABLED_engine_short_names]) {
      en[OVERLAY_PARAM_ENABLED_hide_fps_superscript] = true;
      return "FPS";
   }
```
This line effectively means that "If horizontal mode enabled and long names selected, hide fps superscript and replace with FPS text.  
Doesn't really make sense to me, one of the reasons to implement these two variables was to allow for every engine display setting to be achievable regardless of horizontal or vertical mode.  

Removing this line makes it all work as intended again, except maybe one issue.  
The issue being, this is the default state when none of the above specified options are used:  
image

If this is an issue, maybe we should change these options to `show_fps_superscript` and maybe handle `engine_short_names` differently, what do you think?

JustCryen:
Closed by mistake with a keyboard shortcut.  

@ChrisLane 
> I hadn't seen this but the changes look sensible.
> 
> A couple of minor points to consider for potential improvements:
> 
>     1. FPS superscript still appears for FPS stats in the `fps_metrics` component.
> 
>     2. `hide_engine_names` does not override `engine_short_names` so if the latter appears in config, the former will not take effect.

Regarding your previous points:  
1. I figured out how to get the `hide_fps_superscript` to influence the superscript in the `fps_metrics` component. I'll get that pull request ready after this one is merged.

2. With the above change, it seems like this point is already taken care of.  `hide_engine_names` with or without `engine_short_names` does show just the big FPS text, and hides superscript in the main FPS meter in the process, but also disables the superscript in the `fps_metrics` component.  
Let me know how this should be handled.

JustCryen:
I don't like having fixes ready, waiting in the background and blocked by a README change, thus not being able to push anything forward.  
So I have decided to separate the contribution so I can have both, README changes and code / fixes waiting interdependently for a merge.

PR #1836, opened 2025-10-26
Add cpu_no_label, gpu_no_label, and fps_no_label

samuryan89:
You can currently change the labels for cpu/gpu/fps but not hide them. This change adds options (cpu_no_label, gpu_no_label, and fps_no_label) to hide the labels for a more minimal look. Intended mainly for horizontal layout, but can work in non-horizontal as well.

Before:
image

After:
image

Non-horizontal:
image

### DIFF
--- a/data/MangoHud.conf
+++ b/data/MangoHud.conf
@@ -139,7 +139,7 @@ cpu_stats
 # battery_watt
 # battery_time
 
-Display FPS and frametime
+### Display FPS and frametime
 fps
 # fps_sampling_period=500
 # fps_color_change


flightlessmango:
Instead of adding options, treat an empty string with this behavior

samuryan89:
> Instead of adding options, treat an empty string with this behavior

I can give that a go. I was trying to match the "time_no_label" option.

Are you thinking like:
`cpu_text=""` should make it blank?

Right now that gives the following:
image

flightlessmango:
Yeah that's what I was thinking, it should not have a column like you had designed it

PR #1837, opened 2025-10-26
params: detect sensor availability for "full" overlay preset

matte-schwartz:
gets rid of the extra unused fields for the Z2E (Strix/890m) similar to the other Z-series chips and Vangogh when using the "full" layout

Before:

image

After:

image

### DIFF
--- a/src/amdgpu.cpp
+++ b/src/amdgpu.cpp
@@ -553,6 +553,15 @@ AMDGPU::AMDGPU(std::string pci_dev, uint32_t device_id, uint32_t vendor_id) {
 		}
 	}
 
+	// Probe voltage before the polling thread starts; it shares this FILE*.
+	if (sysfs_nodes.gpu_voltage_soc) {
+		int64_t value = 0;
+		rewind(sysfs_nodes.gpu_voltage_soc);
+		fflush(sysfs_nodes.gpu_voltage_soc);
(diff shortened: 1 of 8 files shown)


flightlessmango:
I wonder if we can figure this out programmatically instead.
If these fields aren't available they *should* report `0xffff` like we check here [[src/amdgpu.cpp:L21](/src/amdgpu.cpp#L21)](https://github.com/flightlessmango/MangoHud/blob/5399bf1a184ce41ff55125e4a24b978392025d24/src/amdgpu.cpp#L73C1-L73C61)

matte-schwartz:
that does sound smarter, I think all of them except for gpu_voltage report as `0xffff` (`gpu_voltage` reports as 0). I can try implementing something based off your idea instead.

flightlessmango:
Seems safe to assume 0 for voltage also means unavailable

matte-schwartz:
sorry, haven't forgotten this. will work on it this week.

matte-schwartz:
reworked this to make it programmatic as suggested, but review would definitely be appreciated here. I've tested this on a few handhelds now (Deck, Go S, Go 2, MSI Claw w/ Lunar Lake) and they seems to work as expected. also found #2070 while working on this, thinking it was my own regression.

Issue #1799, opened 2025-09-08
Hud constantly flickers on Euro Truck Simulator 2 ( radeonsi)


Leopard1907:
**Describe the bug**

Mangohud keeps flickering constantly with Euro Truck Simulator 2 in gameplay, doesn't occur within  in-game menus.

**List relevant hardware/software information**
- Linux Distribution: Arch Linux
- MangoHud version: master
- GPU: RX 7900 XTX
- Mesa version: 25.2.2

**To Reproduce**
1. Install Euro Truck Simulator 2 ( native build )
2. Open the game with Mangohud
3. Drive

**Additional context**

I have tried with and without `AMD_DEBUG=useaco` and there was no change. Issue doesnt occur with Zink.

`MESA_LOADER_DRIVER_OVERRIDE=zink`

```
[2025-09-09 00:39:06.226] [MANGOHUD] [debug] [gl_hud.cpp:134] vendor: AMD, deviceName: AMD Radeon RX 7900 XTX (radeonsi, navi31, ACO, DRM 3.64, 6.16.5-arch1-1)
[2025-09-09 00:39:06.226] [MANGOHUD] [debug] [gl_hud.cpp:157] GL device id: 744C
[2025-09-09 00:39:06.226] [MANGOHUD] [debug] [gl_hud.cpp:159] gpu: AMD Radeon RX 7900 XTX 
[2025-09-09 00:39:06.226] [MANGOHUD] [debug] [gl_renderer.cpp:422] GL version: 4.6 
[2025-09-09 00:39:06.247] [MANGOHUD] [debug] [gl_renderer.cpp:296] glsl_version: 410
```

einhander:
Same here. The problem with 1.55 version of ets2. The 1.53 was fine.

Leopard1907:
@einhander Issue is not reproducible here anymore, seems like a game update fixed it. Can you check as well?

Leopard1907:
Nvm, false alarm. I forgot last time i was opted to use Zink. Still occurs on radeonsi.

Leopard1907:
Fixed by https://github.com/flightlessmango/MangoHud/commit/d4ac646440f8733e97cd83ce4ed64ef122fcd8c0

PR #1840, opened 2025-10-29
GL fixes

Mystral42:
Current MangoHud has some issues when run on OpenGL games that use multiple GL contexts, like many older ddraw-era Wine games end up doing. This fixes them at the cost of effectively never destroying GL resources from old contexts since that's not safe to do at the moment.

The "more proper" fix, I think, would be to wrap the GL context into our own object where we can also keep track of when it's used and destroyed and we can store locally the IDs of the GL resources that are tied to that context. But before I embark into that larger change I thought I'd see how this goes with you :slightly_smiling_face: 

The impetus for all this was to fix Resident Evil 3 Classic on Proton: the overlay would disappear after showing up for just a fraction of a second, and after the first two splash screens the screen would turn entirely black (turns out the latter was because we were destroying random GL objects on the new context which happened to share the same IDs as MangoHud's resources in the old context).

### DIFF
--- a/src/gl/gl.h
+++ b/src/gl/gl.h
@@ -13,8 +13,11 @@ void glXSwapIntervalEXT(void*, void*, int);
 int glXSwapIntervalSGI(int);
 int glXSwapIntervalMESA(unsigned int);
 int glXGetSwapIntervalMESA(void);
+int glXMakeContextCurrent(void*, void*, void*, void*);
 int glXMakeCurrent(void*, void*, void*);
-void* glXGetCurrentContext();
+void *glXGetCurrentContext();
+void *glXGetCurrentDrawable();
+void *glXGetCurrentReadDrawable();
 void *glXCreateContextAttribsARB(void *dpy, void *config,void *share_context, int direct, const int *attrib_list);
 
 void* glXGetProcAddress(const unsigned char*);
--- a/src/gl/gl_hud.cpp
+++ b/src/gl/gl_hud.cpp
@@ -104,18 +104,16 @@ void imgui_init()
     init_cpu_stats(params);
 }
 
-//static
-void imgui_create(void *ctx, const gl_wsi plat)
+void imgui_create(gl_context *ctx, const gl_wsi plat)
 {
-    if (inited)
-        return;
-
+    //SPDLOG_DEBUG("ctx {}", (void *)ctx);
     if (!ctx)
         return;
 
-    imgui_shutdown();
+    if (inited)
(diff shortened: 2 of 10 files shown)


flightlessmango:
> But before I embark into that larger change I thought I'd see how this goes with you

Feel free to refactor as much as you’d like. I’ve been quietly hoping someone would take the time to give this part some love

Mystral42:
Nice! I'll go with those larger changes and see what it looks like. I'll update this PR when I've got something I'm confident with.

Mystral42:
I was planning on wrapping the GL context into our own structure and returning that from e.g. glXCreateContext(), but that plan fell short when I realized that RE3 on Proton uses (wine)gstreamer, which in turn calls glXMakeCurrent() directly (i.e. gstreamer links to libGLX.so). That's a bad problem for that idea, since it would require hooking those kind of calls with some invasive trickery, So I decided to abandon the original plan and just use the current GL context as a key in a map of context structures, not that differently from how the vulkan overlay already works.

I tested this with a few Wine games on 32 and 64 bits, EGL and GLX. It seemed to work well for me but obviously that's not exhaustive testing.

flightlessmango:
From my testing it appears to work good. It also seems to fixed #1849 which is great

src/overlay.cpp
```diff
@@ -696,6 +696,7 @@ void render_imgui(swapchain_stats& data, struct overlay_params& params, ImVec2&
          for (auto& func : HUDElements.ordered_functions){
             if(!real_params->enabled[OVERLAY_PARAM_ENABLED_horizontal] && func.name != "exec")
                ImGui::TableNextRow();
+            //SPDLOG_DEBUG("about to call func name {}", func.name);
```
flightlessmango: Don't forget to remove this remnant

flightlessmango:
Since it's still set as draft, is there something else you'd like to add before merge?

Mystral42:
Oh, no, I just forgot to set it back to normal :sweat_smile:

flightlessmango:
can you squash the commits also

Mystral42:
Sure, all of them into just one?

flightlessmango:
Yes please

Leopard1907:
This pr fixes both https://github.com/flightlessmango/MangoHud/issues/1799 and https://github.com/flightlessmango/MangoHud/issues/1849.

Mystral42:
> Yes please

Done, thanks for the directions :grinning:

PR #1841, opened 2025-10-31
fex: Don't close shm until we successfully open a new one

pac85:
Previously opening the shm file might have failed leaving no shm file open which lead to a SIGSEGV when trying to access the mapped pointer.


Sonicadvance1:
Oops! Thanks for fixing this!

PR #1869, opened 2025-11-27
Request: fix typos

oxyzenQ:
Typos I found (file path, context, suggested correction)

src/shell.cpp :

SPDLOG_DEBUG("Shell: recieved output: {}"..., and comment): "recieved" → should be "received"
Example snippet: SPDLOG_DEBUG("Shell: recieved output: {}" , last_line);

src/app/main.cpp :

Comment: "// make sure that the message recieved is compatible" — "recieved" → "received"

include/nvml.h :

Comment: "timestamps paramter to return the time of each page's retirement." — "paramter" → "parameter"


flightlessmango:
nvml.h is third party so it shouldn't be changed
The rest can be just one commit

oxyzenQ:
> nvml.h is third party so it shouldn't be changed
> The rest can be just one commit

@flightlessmango 
ok, so i just able to 2 commit exclude nvml.h

flightlessmango:
Yeah just make it one commit tho

PR #1886, opened 2025-12-02
vulkan: fix presentModeCount in overlay_CreateSwapchainKHR for vkGetPhysicalDeviceSurfacePresentModesKHR call

andrew-2e128:
presentModeCount was being used incorrectly while also being not initialized resulting in undefined behavior this fix calls vkGetPhysicalDeviceSurfacePresentModesKHR with NULL data first to get the correct presentModeCount.

### DIFF
--- a/src/vulkan.cpp
+++ b/src/vulkan.cpp
@@ -1521,12 +1521,35 @@ static struct overlay_draw *before_present(struct swapchain_data *swapchain_data
    return draw;
 }
 
-static bool IsPresentModeSupported(VkPresentModeKHR targetPresentMode, const std::vector<VkPresentModeKHR>& supportedPresentModes) {
-   for (const auto& mode : supportedPresentModes)
-      if (mode == targetPresentMode)
-         return true;
+static bool is_present_mode_supported(VkPhysicalDevice device, VkSurfaceKHR surface, VkPresentModeKHR targetPresentMode)
+{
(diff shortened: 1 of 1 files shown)


src/vulkan.cpp
```diff
@@ -1556,20 +1556,30 @@ static VkResult overlay_CreateSwapchainKHR(
    (PFN_vkGetPhysicalDeviceSurfacePresentModesKHR) instance_data->vtable.GetInstanceProcAddr(instance_data->instance, "vkGetPhysicalDeviceSurfacePresentModesKHR");
 
    if (fpGetPhysicalDeviceSurfacePresentModesKHR != NULL) {
-      uint32_t presentModeCount;
-      std::vector<VkPresentModeKHR> presentModes(6);
-      VkResult result = fpGetPhysicalDeviceSurfacePresentModesKHR(device_data->physical_device, pCreateInfo->surface, &presentModeCount, presentModes.data());
+      uint32_t presentModeCount = 0;
+      VkResult result;
+      result = fpGetPhysicalDeviceSurfacePresentModesKHR(device_data->physical_device, pCreateInfo->surface, &presentModeCount, NULL);
+      if (result != VK_SUCCESS) {
```
flightlessmango: If it fails we probably shouldn't just guess the amount and continue

PR #1888, opened 2025-12-03
Add panthor driver support

17314642:
Based on #1788, cleaned it up and changed vram key from `panthor-resident-memory` to `drm-resident-memory` because former is only for under-the-hood objects, not related to the games


HeyMeco:
Bump this would be very useful to have

PR #1894, opened 2025-12-06
Add RAM temperature for DDR5 with SPD5118

Henry-ZHR:
Currently one need to enable `ram` to see `ram_temp`. Maybe renaming `ram` -> `ram_usage` and making `ram` like `*pu_stats` would be better?

### DIFF
--- a/README.md
+++ b/README.md
@@ -438,6 +438,7 @@ Parameters that are enabled by default have to be explicitly disabled. These (cu
 | `procmem`<br>`procmem_shared`, `procmem_virt`| Displays process' memory usage: resident, shared and/or virtual. `procmem` (resident) also toggles others off if disabled |
 | `proc_vram`                        | Display process' VRAM usage                                                           |
 | `ram`<br>`vram`                    | Display system RAM/VRAM usage                                                         |
+| `ram_temp`                         | Display RAM temperature (only supports DDR5 with `spd5118` driver)                    |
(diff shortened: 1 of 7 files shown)


Henry-ZHR:
There are temperature drivers for DDR4 etc., but I don't have such device and can't test that

flightlessmango:
This looks good. But it would be nice to leave it open for adding new modules like using the find_input that we do in https://github.com/flightlessmango/MangoHud/blob/d572ba64471223675f81906e9af25d3dd1ede101/src/cpu.cpp#L576

Henry-ZHR:
screenshot

At least for `spd5118`, there is only one `input` so we don't need to find it by `label` (and there is no `label`; also see https://docs.kernel.org/hwmon/spd5118.html)

One should be able to add support for more drivers by adding comparison at https://github.com/flightlessmango/MangoHud/pull/1894/files#diff-da561dadd9636bfa3b09ddee1fbba863f8449a90052d6966e81b6b20780b922bR46 and optionally using something like `find_input`

Henry-ZHR:
Some of the code was copied from other files and looks like there is redundant `.c_str()` too :)

https://github.com/flightlessmango/MangoHud/blob/d572ba64471223675f81906e9af25d3dd1ede101/src/overlay.cpp#L874

flightlessmango:
Squash commits and I think we're fine

Henry-ZHR:
> Squash commits and I think we're fine

Done

PR #1930, opened 2026-01-02
cpu: priority matching for cpu temp

fxzzi:
before, mangohud would loop through hwmon's in order, and use the first one it found that had a valid temperature sensor that we were looking for.

however, in some cases, the hwmon for a worse sensor can be read first, and therefore used, over a more accurate cpu temp sensor. On my system:

```
$ for n in /sys/class/hwmon/hwmon*/name; do echo "$n: $(cat "$n")"; done
/sys/class/hwmon/hwmon0/name: nvme
/sys/class/hwmon/hwmon1/name: nvme
/sys/class/hwmon/hwmon2/name: nct6798
/sys/class/hwmon/hwmon3/name: k10temp
/sys/class/hwmon/hwmon4/name: zenergy
/sys/class/hwmon/hwmon5/name: asusec
/sys/class/hwmon/hwmon6/name: asus
```

the nct temp sensor would be picked first, and this sensor has very bad accuracy to my real temperature, I've seen it showing upwards of 10C lower than the CPU real temperature, better reported by k10temp.

so, implement a priority system when looping through hwmon.

### DIFF
--- a/src/cpu.cpp
+++ b/src/cpu.cpp
@@ -565,58 +565,67 @@ bool CPUStats::GetCpuFile() {
     std::string name, path, input;
     std::string hwmon = "/sys/class/hwmon/";
     std::smatch match;
+    int best_priority = -1;
+    auto const& custom_sensor_name = get_params()->cpu_custom_temp_sensor["hwmon_name"];
+    auto const& custom_sensor_input = get_params()->cpu_custom_temp_sensor["hwmon_input"];
 
     auto dirs = ls(hwmon.c_str());
     for (auto& dir : dirs) {
-        path = hwmon + dir;
-        name = read_line(path + "/name");
-        SPDLOG_DEBUG("hwmon: sensor name: {}", name);
+        const std::string current_path = hwmon + dir;
+        std::string current_input;
+        int priority = -1;
+        
+        const auto name = read_line(current_path + "/name");
+        SPDLOG_DEBUG("hwmon:  sensor name: {}", name);
 
-        std::map<std::string, std::string> custom_sensor = get_params()->cpu_custom_temp_sensor;
-
-        if (!custom_sensor["hwmon_name"].empty() && !custom_sensor["hwmon_input"].empty()) {
-            if (name != custom_sensor["hwmon_name"])
-                continue;
-
-            find_fallback_input(path, custom_sensor["hwmon_input"].c_str(), input);
+        if (name == custom_sensor_name && !custom_sensor_input.empty()) {
+            find_fallback_input(current_path, custom_sensor_input.c_str(), input);
+            // setting a custom sensor should always have the highest prio
+            priority = 100;
+            // also shouldn't need to loop through the rest
             break;
         } else if (name == "coretemp") {
-            find_input(path, "temp", input, "Package id 0");
-            break;
+            if (find_input(current_path, "temp", current_input, "Package id 0"))
+                priority = 10;
         } else if ((name == "zenpower" || name == "k10temp")) {
-            if (!find_input(path, "temp", input, "Tdie"))
-                find_input(path, "temp", input, "Tctl");
-            break;
+            if (find_input(current_path, "temp", current_input, "Tdie") ||
+                find_input(current_path, "temp", current_input, "Tctl"))
+                priority = 9;
         } else if (name == "atk0110") {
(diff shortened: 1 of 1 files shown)


soerengrunewald (approved):
LGTM, in fact it should work even better than the original solution. Since the current implementation does not always check if `find_input` is actually successful, while this does. So we can avoid the situation where we think we have found a valid sensor, but actually have not, therefore :+1: from me.

fxzzi:
rebased on top of main. Commit d24ddd1302b4d8914f849ff3f5f1aa59ce72be51 conflicted but I think I got it working again

src/cpu.cpp
```diff
@@ -565,58 +565,70 @@ bool CPUStats::GetCpuFile() {
...
-            find_fallback_input(path, custom_sensor["hwmon_input"].c_str(), input);
+            find_fallback_input(current_path, custom_sensor["hwmon_input"].c_str(), input);
+            // setting a custom sensor should always have the highest prio
+            priority = 100;
+            // also shouldn't need to loop through the rest
             break;
         } else if (name == "coretemp") {
-            find_input(path, "temp", input, "Package id 0");
-            break;
+            if (find_input(path, "temp", input, "Package id 0"))
```
soerengrunewald: We have to use `current_...` otherwise this is not taken into consideration
```suggestion
            if (find_input(current_path, "temp", current_input, "Package id 0"))
```

src/cpu.cpp
```diff
@@ -565,58 +565,70 @@ bool CPUStats::GetCpuFile() {
...
+                priority = 10;
         } else if ((name == "zenpower" || name == "k10temp")) {
-            if (!find_input(path, "temp", input, "Tdie"))
-                find_input(path, "temp", input, "Tctl");
-            break;
+            if (find_input(path, "temp", input, "Tdie") || 
+                find_input(path, "temp", input, "Tctl"))
```
soerengrunewald: Same here, without `current_path` we will try and empty `path` or the last previously found path.
```suggestion
            if (find_input(current_path, "temp", current_input, "Tdie") || 
                find_input(current_path, "temp", current_input, "Tctl"))
```

src/cpu.cpp
```diff
@@ -565,58 +565,70 @@ bool CPUStats::GetCpuFile() {
...
+        SPDLOG_DEBUG("hwmon:  sensor name: {}", name);
 
         std::map<std::string, std::string> custom_sensor = get_params()->cpu_custom_temp_sensor;
 
         if (!custom_sensor["hwmon_name"].empty() && !custom_sensor["hwmon_input"].empty()) {
             if (name != custom_sensor["hwmon_name"])
                 continue;
 
-            find_fallback_input(path, custom_sensor["hwmon_input"].c_str(), input);
+            find_fallback_input(current_path, custom_sensor["hwmon_input"].c_str(), input);
```
soerengrunewald: We actually don't need to do this on every iteration. Instead we should query `cpu_custom_temp_sensor` only once before the loop.
```diff
+   auto const& custom_sensor_name = get_params()->cpu_custom_temp_sensor["hwmon_name"];
+   auto const& custom_sensor_input = get_params()->cpu_custom_temp_sensor["hwmon_input"];

    auto dirs = ls(hwmon.c_str());
    for (auto& dir : dirs) {
```
And later just use them...



```diff
-        std::map<std::string, std::string> custom_sensor = get_params()->cpu_custom_temp_sensor;
-
-        if (!custom_sensor["hwmon_name"].empty() && !custom_sensor["hwmon_input"].empty()) {
-            if (name != custom_sensor["hwmon_name"])
-                continue;
-
-            find_fallback_input(path, custom_sensor["hwmon_input"].c_str(), input);
-            find_fallback_input(current_path, custom_sensor["hwmon_input"].c_str(), input);
+        // I think we can drop the empty check here, since an empty "custom_sensor_name" name would not match "name"
+        if (name == custom_sensor_name and not custom_sensor_input.empty()) {
+            find_fallback_input(current_path, custom_sensor_input.c_str(), input);
```

soerengrunewald (changes requested):
After actually trying the code, I notice some issues that need to be changed first.

fxzzi:
thanks again for the review and recommendations. I have now implemented those changes, as well as rebased on the latest master.

I kept the empty check since it shouldn't hurt.

Issue #1916, opened 2025-12-23
ImGui fails to draw a label after logging is complete?


QuasiStellar:
Logging completion causes the application to crash.
```
[2025-12-23 17:44:32.045] [MANGOHUD] [info] [logging.cpp:79] /my_log_folder/java_2025-12-23_17-44-01_summary.csv
java: ../MangoHud/subprojects/imgui-1.91.6/imgui.cpp:10783: bool ImGui::ItemAdd(const ImRect&, ImGuiID, const ImRect*, ImGuiItemFlags): Assertion `id != window->ID && "Cannot have an empty ID at the root of a window. If you need an empty label, use ## and read the FAQ about how the ID Stack works!"' failed.
```

**List relevant hardware/software information**
MangoHUD Version: v0.8.2-32-gf00a5f2
Operating System: CachyOS Linux 
KDE Plasma Version: 6.5.4
KDE Frameworks Version: 6.21.0
Qt Version: 6.10.1
Kernel Version: 6.18.2-2-cachyos (64-bit)
Graphics Platform: Wayland
Graphics Processor: NVIDIA GeForce RTX 4080

**To Reproduce**
1. Open goverlay spinning cube simulation (or any other app that supports MangoHUD, e.g. Minecraft)
2. Start logging (Shift_L+F2)
3. Wait for 30 seconds
4. The app crashes
5. The logs are saved but `simulation isn't long enough for precise statistics`

LionHeartP:
```diff
diff --git a/src/overlay.cpp b/src/overlay.cpp
index 2e6f837..f6cb4a7 100644
--- a/src/overlay.cpp
+++ b/src/overlay.cpp
@@ -602,9 +602,9 @@ static void render_benchmark(swapchain_stats& data, const struct overlay_params&
    ImGui::PushStyleColor(ImGuiCol_FrameBg, ImVec4(0.0, 0.0, 0.0, alpha / params.background_alpha));
    ImGui::Dummy(ImVec2(0.0f, 8.0f));
    if (params.enabled[OVERLAY_PARAM_ENABLED_histogram])
-      ImGui::PlotHistogram("", benchmark.fps_data.data(), benchmark.fps_data.size(), 0, "", 0.0f, max + 10, ImVec2(ImGui::GetContentRegionAvail().x, 50));
[...]

PR #1932, opened 2026-01-03
fix ImGui assertion crash in overlay graphs

LionHeartP:
Update ImGui graph calls to include unique IDs to prevent crashes with ImGui 1.91.6+.

In ImGui 1.91.6, passing an empty label ("") to PlotLines or PlotHistogram at the root of a window triggers an assertion failure: `id != window->ID`. 
This occurs because an empty string results in a hash collision with the window ID.

Added hidden IDs to ensure unique hashing without changing the UI.
Relevant ImGui documentation link: https://github.com/ocornut/imgui/blob/master/docs/FAQ.md#q-about-the-id-stack-system

Fixes: #1916


soerengrunewald (approved):
I have only two remarks.
- I would be nice if the description you added for the PR, would be the actual commit message, so one can understand the commit in one year from now (eg. when using git-blame).
- May also add some link to the [ImGui documentation](https://github.com/ocornut/imgui/blob/master/docs/FAQ.md#q-about-the-id-stack-system) in the commit message, so it is easier to understand how you came up with the labels.

Otherwise LGTM :+1:

PR #1945, opened 2026-01-15
ram_temp: add jc42 for DDR4

morph027:
Screenshot From 2026-01-15 20-00-01

Steps taken (on my Gigabyte Aorus B550 Pro V2 with G.Skill Ripjaws F4-3600C14Q @ Samsung B-Dies):

* load module: `modprobe jc42`
* verify:

```bash
❯ grep jc42 /sys/class/hwmon/hwmon*/name
/sys/class/hwmon/hwmon8/name:jc42
/sys/class/hwmon/hwmon9/name:jc42
❯ cat /sys/class/hwmon/hwmon8/temp1_input
28437
```

On my board, `sudo i2cdetect -l` does not list any SMBus for the sensors, so i needed to perform the following step (might be optional):

* add `acpi_enforce_resources=lax` to kernel commandline and reboot

### DIFF
--- a/src/memory.cpp
+++ b/src/memory.cpp
@@ -43,7 +43,8 @@ void update_mem_temp() {
         std::string path = "/sys/class/hwmon/";
         auto dirs = ls(path.c_str(), "hwmon", LS_DIRS);
         for (auto &dir : dirs) {
-            if (read_line(path + dir + "/name") == "spd5118" or read_line(path + dir + "/name") == "jc42")
+            const auto name = read_line(path + dir + "/name");
+            if (name == "spd5118" || name == "jc42")
                 mem_temp_files.emplace_back(path + dir + "/temp1_input");
         }
         if (mem_temp_files.empty())


pallaswept:
> * add `acpi_enforce_resources=lax` to kernel commandline and reboot

This should not be necessary, the jc42 driver does not use any ACPI resources. It talks to the memory over i2c, but perhaps there's an issue with the i2c driver on your motherboard. I'd have to see the error message generated when you're not running that param, to know. I deal with this driver quite a lot, and I've never once seen a board that needs this param for jc42 to load. In any case, the param is required when the motherboard reserves ACPI memory and the driver will not load because it wants to use them, so it's always motherboard-specific whether it is needed.

This module was updated in November '24, so in most cases with a kernel from the last year or so, it will load automatically. Most users with DDR4 ram have working temperature sensors already and don't even know it :)

Anyway thanks OP! I saw mention of this config option in a goverlay commit and was like "why no jc42? I might go fix that" and here you are beat me to it :)

morph027:
The module itself does load indeed, but i2c is missing the buses. I'm running Fedora kernel 6.18 with jc42@rhelversion 10.99 🤷‍♀️ My board also has some other quirks, so probably it's the one to blame. Googling around for `acpi_enforce_resources=lax` also shows a lot more people requiring this for other boards too.

But anyhow, great to hear that it should work out of the box. Just left the comment to make sure future users are able to use it if they stumble upon the same.

src/memory.cpp
```diff
@@ -43,7 +43,7 @@ void update_mem_temp() {
         std::string path = "/sys/class/hwmon/";
         auto dirs = ls(path.c_str(), "hwmon", LS_DIRS);
         for (auto &dir : dirs) {
-            if (read_line(path + dir + "/name") == "spd5118")
+            if (read_line(path + dir + "/name") == "spd5118" or read_line(path + dir + "/name") == "jc42")
```
Henry-ZHR: Every call to `read_line` reads the file (although it's in RAM)

We should read once and then compare all instead

morph027: Yes, tried it but i've never been into C++ and did not figured it out how to do.

I was trying something like a list comprehension in python `if .... in ["spd5118", "jc42"]`.

Help is appreciated.

PR #1965, opened 2026-02-15
vulkan: recognize SDL GPU API

RevengeRip:
From [SDL Wiki](https://wiki.libsdl.org/SDL3/CategoryGPU):

> The GPU API offers a cross-platform way for apps to talk to modern graphics hardware. It offers both 3D graphics and compute support, in the style of Metal, Vulkan, and Direct3D 12.

изображение

### DIFF
--- a/src/overlay.h
+++ b/src/overlay.h
@@ -36,7 +36,8 @@ enum EngineTypes
    FERAL3D,
    TOGL,
 
-   GAMESCOPE
+   GAMESCOPE,
+   SDL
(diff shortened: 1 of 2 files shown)


flightlessmango:
I feel like SDLGPU is a bit awkward, maybe always display it as SDL ?

RevengeRip:
> I feel like SDLGPU is a bit awkward, maybe always display it as SDL ?

Done
изображение

PR #1969, opened 2026-02-19
feat: add a set-fps-limit control command

farnoy:
Preserve control socket if the reloaded config does not change its path. This lets clients keep the connection across preset changes or config reloads.

Update the fps_limiter to directly store the current target, supporting runtime values.

I want to upstream this for use in [mangochill](https://github.com/farnoy/mangochill)

### DIFF
--- a/control/src/control/__init__.py
+++ b/control/src/control/__init__.py
@@ -202,6 +202,9 @@ def control(args):
         conn.send(bytearray(':hud;', 'utf-8'))
     elif args.cmd == 'toggle-fcat':
         conn.send(bytearray(':fcat;', 'utf-8'))
+    elif args.cmd == 'set-fps-limit':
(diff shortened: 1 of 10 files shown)


flightlessmango:
The main reason I’m hesitant to merge this right now is that we’re in the middle of a fairly major architectural shift toward a client/server model, and this change likely wouldn’t be trivially portable to that design.

One option might be to target this toward the server2 branch instead, where most of that work is happening, but I’d like to hear your thoughts first.

If it helps, I’m also happy to talk through how this could fit into the new design, either here in the PR or on Discord, whichever you prefer.

farnoy:
I can rework/retarget the change if needed. What's the timeline on server2 and are there any docs on what's changing?

I could just wait until you're ready to promote server2 as 0.9 or the new mainline.

PR #1987, opened 2026-03-12
params: add fps limit from MANGOHUD_FPS_LIMIT to front of limits list

### LINKED ISSUE #1962: [Feature Request] Allow `fps_limit` limit to be overriden by `MANGOHUD_FPS_LIMIT⁩` environment variable.
At the moment the `fps_limit` option in MangoHud.conf seems to override the `MANGOHUD_FPS_LIMIT⁩` environment variable.

I think it would be more useful, if it was the other way around, so that the user has a convenient way to override the FPS limit on a per-game basis, without having to create a custom MangoHud.conf file for every game that requires a different FPS limit.

Tau5:
When MANGOHUD_FPS_LIMIT is defined it will add that limit to the front of the vector with the fps limits so that it's the default one.
This will allow to use MANGOHUD_FPS_LIMIT even if fps_limit is defined in Mangohud.conf and will also allow the use of the toggle fps limit to use the other limits instead of overwriting them

Fixes #1962

### DIFF
--- a/src/overlay_params.cpp
+++ b/src/overlay_params.cpp
@@ -1088,20 +1088,14 @@ parse_overlay_config(struct overlay_params *params,
                                 );
 
    // check if user specified an env for fps limiter instead
-   if (params->fps_limit.size() == 0 ||
-      (params->fps_limit.size() == 1 && params->fps_limit[0] == 0))
+   const char *fps_limit_env = getenv("MANGOHUD_FPS_LIMIT");
+   if (fps_limit_env)
    {
-      const char *env = getenv("MANGOHUD_FPS_LIMIT");
-      if (env)
-      {
-         try {
-            int fps = std::stof(env);
-            if (params->fps_limit.size() == 0)
-               params->fps_limit.push_back(fps);
-            else
-               params->fps_limit[0] = fps;
-         } catch(...) {}
-      }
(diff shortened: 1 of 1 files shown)


flightlessmango:
Does this actually address https://github.com/flightlessmango/MangoHud/issues/1962 ? Isn't the idea to fully override, not prepend to the front? This is progress anyway, but I don't know if we can say it fixes the issue

Tau5:
I thought it would be better this way than just overriding it. By default it provides the same behaviour but it provides more flexibility.

But if you think it would be better to just override it, it's not hard to change it

flightlessmango:
I think this is good. My only concern is saying that it fixes the issue when the author describes it a bit differently

Tau5:
So, should I remove the mention about fixing the issue?

flightlessmango:
Either remove the mention or confirm with the issue author that this resolves it

CNR0706:
Author of #1962 here. I'm more than happy with this PR.

IMO this approach is superior to completely overriding the configured FPS limits, as it allows for additional flexibility (switching FPS caps at runtime).

Feel free to close #1962 once this is merged and working.

CNR0706:
Works great. Thanks a lot!
afbeelding

PR #1992, opened 2026-03-16
Guard user specified Vulkan present mode against dynamic changes

silvanshade:
This PR fixes a problem where a user-specified Vulkan present mode override (either through `vsync=` or the new `vulkan_present_mode=`) is forgotten and ignored if the present mode is changed dynamically after original creation.

This happens almost always with `dxvk` and `vkd3d` or anything using swapchain maintenance to change present modes.

I suspect the reason it may not have been noticed before (or maybe it has been, I haven't searched the issues) is because prior to #1988 the present mode reporting for Vulkan was usually incorrect since it did not account for dynamic changes.

So it would have appeared that the override worked, since MangoHud would report that initially at creation, but then it would be changed and the reporting never updated to reflect the fact that the override wasn't doing anything.

### DIFF
--- a/src/overlay_params.cpp
+++ b/src/overlay_params.cpp
@@ -1209,6 +1209,8 @@ parse_overlay_config(struct overlay_params *params,
       if (parse_vulkan_present_mode_name(params->vulkan_present_mode, present_mode)) {
          params->m_vulkan_present_mode = present_mode;
       }
+   } else if ((int)params->vsync != -1) {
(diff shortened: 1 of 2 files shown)


flightlessmango:
Heads up: we're removing the VkSwapchainPresentModeInfoEXT patching from this PR in 54ae1cfaaa7a8612e0746211d05648d337ac6064

As discussed in #2056, we've found that modifying the application's VkSwapchainPresentModeInfoEXT data can cause compatibility issues, so we'd rather not touch it.

PR #2000, opened 2026-03-27
Add gpu_hotspot_delta metric (junction - edge temperature)

Geoknyda:
This PR adds a new optional metric `gpu_hotspot_delta`, which displays the difference between GPU junction (hotspot) temperature and edge temperature.

This is particularly useful for AMD GPUs (RDNA2/RDNA3), where hotspot delta is commonly used to assess cooling performance, mounting pressure, and thermal paste condition.

Key characteristics:
- Uses existing metrics (no additional sensor polling)
- Lightweight and low overhead
- Disabled by default
- Supports both Celsius and Fahrenheit modes
- Integrated consistently with existing GPU metrics

Example:
GPU Temp: 72°C
GPU Junction: 94°C
dT: 22°C

This provides users with immediate visibility into hotspot spread without requiring manual calculation.

### DIFF
--- a/src/hud_elements.cpp
+++ b/src/hud_elements.cpp
@@ -344,6 +344,36 @@ void HudElements::gpu_stats(){
                 ImGui::PopFont();
             }
 
+            if (gpu->metrics.junction_temp > -1 &&
+                gpu->metrics.temp > -1 &&
+                HUDElements.params->enabled[OVERLAY_PARAM_ENABLED_gpu_hotspot_delta]) {
+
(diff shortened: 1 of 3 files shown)


flightlessmango:
Thanks for the PR and the work put into it.

I’m going to decline this for now because I don’t want to add hardcoded handling for this kind of value transformation. I’d rather support a more general config approach where users can do basic math on values, which would cover this use case and a broader range of similar ones.

Because of that, I don’t think it makes sense to merge this in its current form.

PR #2019, opened 2026-04-17
Implement detection of proton 11 version

Etaash-mathamsetty:


flightlessmango:
Probably just need to ifdef linux around it to fix mingw

Etaash-mathamsetty:
> Probably just need to ifdef linux around it to fix mingw

Fixed

Issue #1566, opened 2025-02-13
read_cfg setting causes items to show up in HUD twice


PatrickSzela:
**Describe the bug**
Since the `0.8.0` release, using the `read_cfg` setting in the `MANGOHUD_CONFIG` environmental variable causes (some) presets to contain duplicated items

**List relevant hardware/software information**
- Linux Distribution: Arch Linux
- MangoHud version: 0.8.0
- GPU: Radeon RX 6750 XT

**To Reproduce**
1. Execute `MANGOHUD=1 MANGOHUD_CONFIG=read_cfg vkcube`
2. Switch to a different preset with `Shift+F10` keybind, either 1 (fps only) or 2 (horizontal view)
3. Notice the duplicated items in the HUD

**Expected behavior**
Items should not be duplicated when using the `read_cfg` setting with `MANGOHUD_CONFIG` environmental variable

**Screenshots**
- Preset 1 (fps only)
![Image](https://github.com/user-attachments/assets/0e990449-37b2-47eb-bbcc-c51bd9ca8884)
- Preset 2 (horizontal view)
![Image](https://github.com/user-attachments/assets/69ca1321-aba4-4411-9476-09a76e2a817e)

**Additional context**
When using default MangoHUD configuration (so no custom .conf files), only presets 1 (fps only) or 2 (horizontal view) are affected by this bug - see pictures above.

With my own .conf files, all my preset seems to be affected. To save up some space, I'm including only one of my presets:
- `presets.conf`:
```
[preset 5]
legacy_layout=false
horizontal
horizontal_stretch=false
hud_no_margin
font_size=14
text_outline_thickness=1.0
time_format=%R
time_no_label
time
fps
frametime=false
```
- `MangoHud.conf`:
```
preset=5
```
Additionally, this issue seems to persist even if paired with other settings - so far I've tested it with `no_display` and `help`

PatrickSzela:
After some more testing, it seems like this issue only occurs for presets that use `legacy_layout=false`

Jimbleton115:
running into this issue as well, although it seems to not happen if you don't use a presets.conf file; if I put a layout into the MangoHud.conf file directly, it works as usual with no duplicated entries.

astralaster:
Can confirm this bug. Mangohud version 0.8.4 still has it. And it does not duplicate with legacy_layout=1 as said above.

flightlessmango:
c9f7b36404017b74f0105f2378d638dab09c852f should be resolved here

PR #2062, opened 2026-06-02
Fix read_cfg setting causes items to show up in HUD twice for presets using legacy_layout=false

astralaster:
This change fixes this issue: https://github.com/flightlessmango/MangoHud/issues/1566

This is my first contribution and I have done some basic testing if mangohud still works as expected after this change, but I might have missed something.


flightlessmango:
c9f7b36404017b74f0105f2378d638dab09c852f resolved here

PR #2064, opened 2026-06-05
pkgbuild: Updated for mangohud-next

Billli11:
~~vkroots install header by default.~~

Found this option in `meson install`

```
--skip-subprojects [SKIP_SUBPROJECTS]  Do not install files from given subprojects. (Since 0.58.0)
```


flightlessmango:
vkroots header installation has been fixed, you can remove `skip-subprojects`

Billli11:
Done. But vkroot still need some clean up. #2066

```
% pacman -Ql mangohud lib32-mangohud|grep vkroot
mangohud /usr/lib/pkgconfig/vkroots.pc
lib32-mangohud /usr/lib32/pkgconfig/vkroots.pc
```

PR #2066, opened 2026-06-06
vkroots: do not install pkgconfig

Billli11:
This is also not needed.


flightlessmango:
already done d6ebab4d949a9837e932fb24a046255a5c018180

PR #2071, opened 2026-06-10
MangoHud-Next: compilation under elogind

neptuwunium:
Allows for mangohud-next to be compiled on systems with elogind as a logind provider instead of systemd when the system does not have systemd (i.e. openrc.)

Reasoning for adding IPC module to server, metrics, render and imgui, they all consume ipc headers and under elogind the headers are not under the system include dir but instead `elogind/systemd...`.

### DIFF
--- a/mangohud-next/ipc/meson.build
+++ b/mangohud-next/ipc/meson.build
@@ -1,4 +1,4 @@
-systemd_dep = dependency('libsystemd')
+systemd_dep = dependency('libsystemd', 'libelogind')
 
 ipc_inc = include_directories('.')
 
@@ -45,4 +45,5 @@ ipc_lib = static_library(
 ipc_dep = declare_dependency(
   link_with: ipc_lib,
   include_directories: [ipc_inc],
+  dependencies: [systemd_dep]
 )
--- a/mangohud-next/meson.build
+++ b/mangohud-next/meson.build
@@ -8,11 +8,11 @@ inc_common += [
 subdir('utils')
 vkbootstrap_dep = subproject('vk-bootstrap').get_variable('vkbootstrap_dep')
 libdrm_dep = dependency('libdrm')
+subdir('ipc')
 # subdir('render')
 if get_option('with_server')
   subdir('render')
 endif
-subdir('ipc')
 if get_option('with_server')
   subdir('server')
 endif
--- a/mangohud-next/render/imgui/meson.build
+++ b/mangohud-next/render/imgui/meson.build
@@ -3,6 +3,6 @@ subdir('font')
 mangohud_imgui = static_library(
     'mangohud_imgui',
     files('imgui_ctx.cpp'),
-    dependencies: [dearimgui_dep, spdlog_dep, vkbootstrap_dep, implot_dep],
+    dependencies: [dearimgui_dep, spdlog_dep, vkbootstrap_dep, implot_dep, ipc_dep],
     include_directories: inc_common
 )
--- a/mangohud-next/render/meson.build
+++ b/mangohud-next/render/meson.build
@@ -15,7 +15,8 @@ render_lib = static_library(
                    dependency('libdrm'),
                    dependency('gl'),
                    dependency('egl'),
-                   dep_vulkan_utility_libraries],
+                   dep_vulkan_utility_libraries,
+                   ipc_dep],
     link_with: mangohud_imgui,
     include_directories: inc_common
 
--- a/mangohud-next/server/metrics/meson.build
+++ b/mangohud-next/server/metrics/meson.build
@@ -12,7 +12,8 @@ metrics_lib = static_library(
                   vkbootstrap_dep,
                   dearimgui_dep,
                   implot_dep,
-                  vulkan_dep],
+                  vulkan_dep,
+                  ipc_dep],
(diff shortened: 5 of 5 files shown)


mangohud-next/ipc/meson.build
```diff
@@ -1,4 +1,4 @@
-systemd_dep = dependency('libsystemd')
+systemd_dep = dependency('libsystemd', 'libelogind')
```
flightlessmango, line 1: This is incorrect usage, you would have to check if libsystemd was found and then try libelogind

neptuwunium: Incorrect, the usage here already does that.

> The names of the dependency to look up. The dependencies are looked up in the order they are provided here. The first found dependency will then be used.

https://mesonbuild.com/Reference-manual_functions_dependency.html

mangohud-next/ipc/meson.build
```diff
@@ -45,4 +45,5 @@ ipc_lib = static_library(
 ipc_dep = declare_dependency(
   link_with: ipc_lib,
   include_directories: [ipc_inc],
+  dependencies: [systemd_dep]
```
flightlessmango, line 48: Not sure this should be needed

neptuwunium: It tells meson that dependents of ipc_dep need to also depend on systemd_dep, which is needed because of ipc headers relying on systemd headers.

~~Though a more reasonable solution would be to use `systemd_dep.partial_dependency(includes : true)`~~ This does not work as expected.

mangohud-next/ipc/meson.build
flightlessmango: You're right 👍

mangohud-next/ipc/meson.build
flightlessmango: Ideally render doesn't depend on ipc and so doesn't depend on systemd, which I thought it didn't, but that might be by accident

neptuwunium: I can build current HEAD without the ipc_dep fixups and see where the offending includes are happening in a bit.

neptuwunium:
```
In file included from ../mangohud-9999/mangohud-next/server/metrics/metrics.cpp:1:
In file included from ../mangohud-9999/mangohud-next/server/metrics/metrics.h:11:
../mangohud-9999/mangohud-next/server/metrics/../../ipc/ipc.h:2:10: fatal error: 'systemd/sd-bus.h' file not found

In file included from ../mangohud-9999/mangohud-next/render/imgui/imgui_ctx.cpp:5:
In file included from ../mangohud-9999/mangohud-next/render/imgui/imgui_ctx.h:7:
In file included from ../mangohud-9999/mangohud-next/render/imgui/../vulkan_ctx.h:8:
../mangohud-9999/mangohud-next/render/imgui/../../ipc/client.h:11:10: fatal error: 'systemd/sd-bus.h' file not found

In file included from ../mangohud-9999/mangohud-next/render/egl_ctx.cpp:17:
In file included from ../mangohud-9999/mangohud-next/render/imgui/egl.h:7:
In file included from ../mangohud-9999/mangohud-next/render/imgui/imgui_ctx.h:7:
In file included from ../mangohud-9999/mangohud-next/render/imgui/../vulkan_ctx.h:8:
../mangohud-9999/mangohud-next/render/imgui/../../ipc/client.h:11:10: fatal error: 'systemd/sd-bus.h' file not found

In file included from ../mangohud-9999/mangohud-next/render/vulkan_ctx.cpp:4:
In file included from ../mangohud-9999/mangohud-next/render/vulkan_ctx.h:8:
../mangohud-9999/mangohud-next/render/../ipc/client.h:11:10: fatal error: 'systemd/sd-bus.h' file not found
```

this is with only the `mangohud-next/ipc/meson.build` changes

neptuwunium: It depends on client.h for:

```
In file included from ../mangohud-9999/mangohud-next/render/vulkan_ctx.cpp:4:
../mangohud-9999/mangohud-next/render/vulkan_ctx.h:29:45: error: use of undeclared identifier 'Client'
   29 |     void wait_on_semaphores(std::shared_ptr<Client> client);
      |                                             ^~~~~~
../mangohud-9999/mangohud-next/render/vulkan_ctx.h:30:36: error: use of undeclared identifier 'Client'
   30 |     void sync_wait(std::shared_ptr<Client> client);
      |                                    ^~~~~~
```

flightlessmango:
implemented here 4a2473c85c21b005f086b38a07ed5c34e1e2ddd8

neptuwunium:
Unfortunately the renderer still relies on IPC headers which rely on systemd;

```
In file included from mangohud-next/render/vulkan_ctx.cpp:4:
In file included from mangohud-next/render/vulkan_ctx.h:8:
mangohud-next/render/../ipc/client.h:13:10: fatal error: 'systemd/sd-bus.h' file not found
   13 | #include <systemd/sd-bus.h>
      |          ^~~~~~~~~~~~~~~~~~
```

flightlessmango:
should be resolved here 785507cee7d845d22d2e38fca06bb367f73915bd

PR #2077, opened 2026-06-18
next: fix per-client resource leaks in the server

samto6:
The mangohud-next server builds per-client resources on connect and tears them down on disconnect. Over a session of game launches, its open fd count climbs until the server stops serving new clients, and two ownership bugs in that path leak file descriptors. A third problem is masked by the client leak: once the client is freed correctly, nothing keeps the per-renderer Vulkan context alive between connects, so it is rebuilt each time, leaking descriptors the driver never frees on teardown.

Fix 1:
The gbm device and buffer object backing each client's dmabuf were never freed. Their cleanup (`gbm_bo_destroy` and `gbm_device_destroy`) sat in a method named after the struct, `gbmBuffer()`, so it was the default constructor rather than the destructor and ran while the pointers were still null. Each client leaked its gbm render node descriptors. Own the device and buffer object through RAII, in a `gbmAlloc` that holds them in smart pointers with the right deleters, so they are released when the buffer is destroyed.

Fix 2:
The per-client work queue held tasks that owned the client. `send_dmabuf`, `send_config` and `frame_ready` each posted a lambda capturing a `shared_ptr` to the `Client`, and the queue is itself a member of that `Client`, so a queued task was a strong reference from the client back to itself. When a client disconnected with a frame still in flight, the leftover task kept the `Client` alive and nothing it owned was released: its two eventfds, its event loop sources, in-flight fences, and the Vulkan render context it referenced. Change the queue to hold `void(Client*)` tasks and pass in the `Client` pointer the handler already holds.

Fix 3:
The per-renderer Vulkan context was cached in a `weak_ptr`, held strongly only by the connected client, so it was destroyed once the last client referencing it went away and a fresh Vulkan instance and device were built on the next connect. Hold it in a `shared_ptr` so it is built once and shared. While here, drop `VkCtx::gbm_dev`: it was never read, `create_gbm` builds its own device from `phys_fd()`, and with the context now persistent `init_client` would recreate it per connect while the destructor runs once, leaking one each time.

### DIFF
--- a/mangohud-next/render/shared.h
+++ b/mangohud-next/render/shared.h
@@ -93,7 +93,36 @@ struct gbmBuffer {
     uint32_t fourcc = 0;
     uint64_t plane_size = 0;
 
-    gbmBuffer() {
+    gbmBuffer() = default;
+
+    gbmBuffer(const gbmBuffer&) = delete;
+    gbmBuffer& operator=(const gbmBuffer&) = delete;
+
+    gbmBuffer(gbmBuffer&& o) noexcept { *this = std::move(o); }
+
+    gbmBuffer& operator=(gbmBuffer&& o) noexcept {
+        if (this != &o) {
+            if (bo)
+                gbm_bo_destroy(bo);
+            if (dev)
+                gbm_device_destroy(dev);
+
+            dev = o.dev;
+            bo = o.bo;
+            fd = std::move(o.fd);
+            modifier = o.modifier;
+            stride = o.stride;
+            offset = o.offset;
+            fourcc = o.fourcc;
+            plane_size = o.plane_size;
+
+            o.dev = nullptr;
+            o.bo = nullptr;
+        }
+        return *this;
+    }
+
+    ~gbmBuffer() {
         if (bo)
             gbm_bo_destroy(bo);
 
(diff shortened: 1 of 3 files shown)


flightlessmango:
Fix 3: the client being the only thing keeping the vulkan context alive is intended. The context is created and destroyed on demand

mangohud-next/render/shared.h
```diff
@@ -92,15 +121,6 @@ struct gbmBuffer {
 
     uint32_t fourcc = 0;
     uint64_t plane_size = 0;
-
-    gbmBuffer() {
```
flightlessmango, line 96: The leak is real, but the fix could just be correcting this typo by making this `~gbmBuffer()`.

`gbmBuffer` already owns both `dev` and `bo`, and `bo` only needs to be destroyed before `dev`, which the destructor can do directly. Wrapping them in unique_ptr does not seem to buy much here

flightlessmango:
Fix 1: The leak comes from abandoned queued tasks holding shared_ptr of Client. instead of reworking this, we can preserve the original intent by clearing work_q during shutdown.

samto6:
Commit ada323a now uses raw `dev` and `bo` with a `~gbmBuffer()`. Just renaming it doesn't compile because the destructor suppresses the implicit moves and `unique_fd` makes copies deleted, so neither `dmabuf.gbm = {}` nor the resized `buffer` vector builds. The move ctor/assign was added to address this.

5375e7f just clears `work_q` in `stop_and_join` now.

Fix 3 was dropped; the fd growth I saw was the layer loading into the server itself in my test setup, not the context.

PR #2085, opened 2026-07-05
Use next's GPU metrics implementation for GPU_fdinfo class

17314642:
Every driver except for amdgpu and nvidia will now use gpu metrics from mangohud-next

I tested that it works on Intel ARC A770 using both 64-bit and 32-bit builds and mangoapp in gamescope.
proc_vram using fdinfo from mangohud-next for amdgpu also works

I implemented it using opaque pointer technique to not expose any headers from next to legacy and also making as little changes as possible to the mangohud-next's codebase.


flightlessmango:
Could you squash or reorder e16441d0 and a5b77df1 so the amdgpu fdinfo path is not broken on bisect?

```../../src/gpu_fdinfo.cpp: In member function ‘float GPU_fdinfo::amdgpu_helper_get_proc_vram()’:
../../src/gpu_fdinfo.cpp:23:15: error: base operand of ‘->’ has non-pointer type ‘LegacyGPUWrapper’
   23 |     return gpu->get_process_vram_used(pid);
      |               ^~
ninja: build stopped: subcommand failed.```

17314642:
Sure thing, done!

17314642:
@flightlessmango since #2073 is closed (msm metrics are fixed in this pr), this is fully ready!

PR #2089, opened 2026-07-14
fix(amdgpu): force sysfs hwmon only for Vega20/MI50 metrics

johncoffee715:
## Bug

MangoHud 0.8.4 displays incorrect GPU metrics (zeroed or exorbitant values) on **AMD Radeon Pro VII / MI50 (Vega 20, PCI ID 0x66A1)**.

## Root cause

The SMU `gpu_metrics` binary blob (`/sys/bus/pci/devices/0000:05:00.0/gpu_metrics`) returns incorrect values on Vega 20 / gfx906. The `gpu_metrics` file exists and is readable, so `gpu_metrics_is_valid` is set to `true`, but the data inside (temperatures, clocks, power, fan speed) is unreliable.

The code in `amdgpu.cpp` then **overrides** the correct sysfs hwmon values with the incorrect gpu_metrics values:

```cpp
if (gpu_metrics_is_valid) {
    metrics.temp = amdgpu_common_metrics.gpu_temp_c;  // overrides correct hwmon temp
    // ...
}
```

## Fix

Force `gpu_metrics_is_valid = false` so MangoHud reads exclusively from sysfs hwmon — the same data source `amdgpu_top` uses successfully on this hardware.

## Verification

- `amdgpu_top` 0.11.5 reads all metrics correctly via DRM ioctl + sysfs hwmon on this GPU
- With this patch, MangoHud reads from the exact same sysfs hwmon path (`/sys/bus/pci/devices/0000:05:00.0/hwmon/hwmon0/`)
- All metrics remain available via sysfs: `gpu_busy_percent`, `freq1/2_input`, `temp1/2/3_input`, `power1_input/cap`, `fan1_input`, `in0_input`, `mem_info_vram_total/used`
- Only throttle status is lost (gpu_metrics-only field, minor)

## Test environment

- GPU: AMD Radeon Pro VII (Vega 20, PCI ID 0x66A1, VBIOS 113-D1640700-100)
- OS: CachyOS Linux (Arch-based), kernel 6.x
- Driver: amdgpu (open)
- Vulkan: RADV VEGA20
- `amdgpu_top` 0.11.5 confirms all sysfs hwmon sensors return correct values

## Alternative approach

Instead of hardcoding `false`, a config option (`amdgpu_sysfs_only`) or a device ID allowlist could be added. I chose the simple approach because Vega 20 is the only known affected ASIC, and the sysfs hwmon path is strictly more reliable on this hardware.

### DIFF
--- a/contrib/99-powercap-rapl.rules
+++ b/contrib/99-powercap-rapl.rules
@@ -1,0 +1,4 @@
+# Allow video group to read RAPL energy counter for MangoHud CPU power
+# Without this rule, /sys/class/powercap/intel-rapl:0/energy_uj is 0400 root:root
+# and MangoHud cannot display CPU power consumption on Intel CPUs
+SUBSYSTEM=="powercap", KERNEL=="intel-rapl:0", RUN+="/bin/chgrp video /sys/class/powercap/intel-rapl:0/energy_uj", RUN+="/bin/chmod 040 /sys/class/powercap/intel-rapl:0/energy_uj"
--- a/contrib/install-rapl-fix.sh
+++ b/contrib/install-rapl-fix.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+# Install RAPL powercap permissions fix for MangoHud CPU power display
+# Fixes: CPU power (watts) not showing in MangoHud overlay on Intel CPUs
+# Root cause: /sys/class/powercap/intel-rapl:0/energy_uj is 0400 root:root
+set -e
+
+RULE_SRC="$(dirname "$0")/99-powercap-rapl.rules"
+RULE_DST="/etc/udev/rules.d/99-powercap-rapl.rules"
+
+if [ "$(id -u)" -ne 0 ]; then
+    echo "Run with sudo or pkexec"
+    exit 1
+fi
+
+cp "$RULE_SRC" "$RULE_DST"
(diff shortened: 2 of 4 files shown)


johncoffee715:
## Update: Added RAPL debug logging + powercap udev rule

Two additional commits pushed:

### 1. RAPL debug logging (`cpu.cpp`)
Added `SPDLOG_TRACE` to `get_cpu_power_rapl()` so the actual CPU power value is logged:
```
RAPL: power = 81 W (delta=4777271 uJ, dt=58520 us)
```
This helps users verify CPU power is being read correctly.

### 2. Contrib: RAPL powercap udev rule
On Intel CPUs, `/sys/class/powercap/intel-rapl:0/energy_uj` defaults to `0400 root:root`, preventing MangoHud from reading CPU power consumption. The included udev rule (`contrib/99-powercap-rapl.rules`) makes it readable by the `video` group.

Install:
```bash
sudo contrib/install-rapl-fix.sh
```

### Verification
Tested with `MANGOHUD_LOG_LEVEL=trace vkcube`:
```
[trace] RAPL: power = 81 W (delta=4777271 uJ, dt=58520 us)
[trace] RAPL: power = 78 W (delta=39406820 uJ, dt=500434 us)
[trace] RAPL: power = 73 W
...
```

CPU power now displays correctly in the overlay.

PR #2098, opened 2026-07-28
Add frametime coefficient of variation metric

### LINKED ISSUE #2023: [FEATURE REQUEST] Add frametimes relative standard deviation (Coefficient of Variation) measurement
Frametime stability and smoothness is probably better measured with standard deviation based metrics (https://en.wikipedia.org/wiki/Coefficient_of_variation) more than just percentiles.
While percentiles are good for capturing stutters and some instability, they tend to be completely blind to the chaos near the average. Standard deviation divided by the mean could be a great addition as it's standard statistical tool that reflect all variance in all samples, with skew toward outliers. 

One problem is this measure is not common even in advanced tools like presentmon, so people may not be familiar with its implication, but we can present it as 'stability' or other common terms.

If this is within the scope of mangohud, i'm willing to work on it.

OursCodeur:
Closes #2023.

Hello ! This adds `cv` to `fps_metrics`. It reports the population standard deviation divided by mean frametime for the current sample window, as a percentage. Lower values indicate steadier frametimes.

The metric is shown with two decimal places and a percent suffix in both the regular HUD and the post-logging benchmark overlay. The percent suffix remains visible when `hide_fps_superscript` is enabled, since otherwise the value could be mistaken for FPS.

The calculation uses a double-precision two-pass implementation, configurations containing only `cv` skip the percentile copy and sort (the first presentation is excluded because it has no previous presentation timestamp).

README and sample configuration document the new token. Perf-wise at the 10,000-sample cap, CV arithmetic adds roughly 12–14 µs; a CV-only calculation takes approximately 24.6 µs end to end on an i7-13700KF.

I left the benchmark CSV schema unchanged. If CV should be exported there as well, how would you prefer the column to be represented and ordered?

Cheers

mangohud_CV

### DIFF
--- a/README.md
+++ b/README.md
@@ -396,7 +396,7 @@ Parameters that are enabled by default have to be explicitly disabled. These (cu
 | `fps_only`                         | Show FPS only. ***Not meant to be used with other display params***                   |
 | `fps_sampling_period=`             | Time interval between two sampling points for gathering the FPS in milliseconds. Default is `500`   |
 | `fps_value`                        | Choose the break points where `fps_color_change` changes colors between. E.g `60,144`, default is `30,60` |
-| `fps_metrics`                      | Takes a list of decimal values or the value avg, e.g `avg,0.001`                      |
+| `fps_metrics`                      | Takes a list of decimal percentile values, `avg`, or `cv`. CV is the population standard deviation divided by mean frametime, in percent; lower is steadier. E.g. `avg,0.001,cv` |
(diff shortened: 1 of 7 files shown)


zarbis:
I would really love to see this feature merged! I think it's an important step in evolution of performance measurement.
I think CoV is the next logical step after FPS and Percentiles, where each new metric gives us hard numbers to subjective feeling of how good or bad animation quality feels.

PR #2127, opened 2026-08-28
fix: report an absent sensor as -1 rather than 0

GribanovIvan:
junction_temp and memory_temp are declared with a -1 default and the HUD only draws them when the value is greater than -1, but the constructor initialises both to 0 and so defeats both. Any GPU without those sensors therefore renders a steady "0 C" instead of hiding the row - an APU exposing only an edge sensor shows 0 C junction forever.

Initialise them to -1 so the declared default and the draw condition agree.

### DIFF
--- a/src/gpu_metrics_util.h
+++ b/src/gpu_metrics_util.h
@@ -25,7 +25,7 @@ struct gpu_metrics {
     bool fan_rpm;
 
     gpu_metrics()
-        : load(0), temp(0), junction_temp(0), memory_temp(0),
+        : load(0), temp(0), junction_temp(-1), memory_temp(-1),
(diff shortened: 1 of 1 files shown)


flightlessmango:
The cleaner solution might be to remove the constructor entirely and set defaults

flightlessmango:
this can be one commit

GribanovIvan:
If squash merging is enabled for the repo, Squash and merge folds them into one at merge time: https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/configuring-commit-squashing-for-pull-requests
Or do you actually want me to force-push the branch as a single commit?

PR #2135, opened 2026-09-06
Search for batteries by checking directories under /sys/class/power_supply/

PourOutYourPowerAndLove:
Additionally, check if the sysfs attribute actually returns a number which there is maybe a better way of doing it, but it is necessary because on my system, the charge_now and charge_full attributes return "No data available", in which case just ends up returning battery %NaN. 

Do the same when checking the remaining battery life as well

### DIFF
--- a/src/battery.cpp
+++ b/src/battery.cpp
@@ -7,19 +7,32 @@ using namespace std;
 
 void BatteryStats::numBattery() {
     int batteryCount = 0;
-    if (!fs::exists("/sys/class/power_supply/")) {
-         batteryCount = 0;
+    
+    if (fs::exists("/sys/class/power_supply/")) {
+    	fs::path path("/sys/class/power_supply/");
+    	for (auto& p : fs::directory_iterator(path)) {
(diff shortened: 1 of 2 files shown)


17314642:
personally, I would refactor your code as shown below. It's cleaner looking, avoids long lines and repetition of `charge` and `energy` keywords

```c++
void BatteryStats::numBattery() {
    const fs::path powerSupplyPath("/sys/class/power_supply");
    int batteryCount = 0;

    if (!fs::exists(powerSupplyPath)) {
        return;
    }

    for (const fs::directory_entry& p : fs::directory_iterator(powerSupplyPath)) {
        const std::string curDir = p.path();

        for (const std::string s : {"/charge", "/energy"}) {
            if (!fs::exists(curDir + s + "_now") || !fs::exists(curDir + s + "_full")) {
                continue;
            }

            //Test if path contains the stuff we use if it is a battery
            battPath[batteryCount] = curDir;
            battType[batteryCount] = "/none";

            std::ifstream ifs(curDir + s + "_now");
            std::string line;

            std::getline(ifs, line);

            std::istringstream ss(line);

            if (int val; ss >> val) {
                battType[batteryCount] = s;
                batteryCount += 1;
                break;
            }
        }
    }

    batt_count = batteryCount;
    batt_check = true;
}
```

17314642:
And maybe change size of battPath and battType to 10, in case some devices have more than 2 battery sources. It's a band-aid approach, but it avoids changing more code if we were to change it to std::vector

PR #2136, opened 2026-09-06
Fix reporting CPU temperatures when thermal zones are used

PourOutYourPowerAndLove:
The regex pattern match was wrong and wouldn't pick up on the sysfs directories because they are named like cpu_0_0_0_thermal_2 and so on

ez fix


17314642:
This breaks cpu temperature on rockchip boards, and possibly some others

Instead of removing old regex, consider checking with both regexes

Image below is from a year ago, back when I was adding cpu temperature support, and you can see that in this case, cpu temperature sensor is called `cpu_thermal`, while newer rockchip boards have `cpu0_thermal` and so on
646878290-cd5df63a-9def-421b-b5c6-1f79c43aa00e

17314642:
646878839-f7bb992f-fde8-473a-bb0d-8c30e55ccb31

Issue #1950, opened 2026-01-19
Feature: Show only temperatures


michaelneverwins:
Currently, as far as I can tell, there is no way to display CPU or GPU temperature without also displaying the usage/load percentage. In other words, `cpu_temp` and `gpu_temp` don't do anything unless also using `cpu_stats` and `gpu_stats`, respectively, each of which causes MangoHud to show a percentage.

I'd like to be able to show only temperatures, for two reasons:
1. Frankly, I really don't _care_ about the CPU/GPU usage as a percentage. What I want to know is whether my hardware is running too hot. Generally one thing will correlate to the other, but the temperature is what I'd rather see at a glance, and that would be easier with fewer numbers.
2. MangoHud can't even get the GPU usage in one of my PCs (which has a Radeon HD 7950, and based on information in other bug reports like #233, I've always assumed there's no fix for that). But MangoHud _can_ measure the temperature of this GPU, which is what I really want to see anyway. Having to see GPU usage stuck at 0% in order to see the temperature is just a bit annoying, especially when I don't really care to see the usage anyway. I'd rather just hide it. I know this is an extremely specific situation, but I'm sure I'm not the only one with a card for which MangoHud can see temperature and not usage.

drewboardman:
This is still an open issue. I've submitted a PR here: https://github.com/flightlessmango/MangoHud/pull/2138

Please comment/review.

PR #2138, opened 2026-09-06
hud: add temperature-only CPU and GPU options

drewboardman:
## Summary

Add `cpu_temp_only` and `gpu_temp_only` configuration options.

This was submitted as an issue on jan 2026: https://github.com/flightlessmango/MangoHud/issues/1950

I added these in a backwards-compatible way (they are new options):

```ini
cpu_temp_only
gpu_temp_only
```

However, I think at some point it may make sense to look into having `gpu_temp` and `cpu_temp` be "temp only". The current implementation is confusing. I'd suggest something like `cpu_temp_full` to show the % load as well. Just a suggestion.

I've refactored the shared behavior into a helper.

These options display the corresponding CPU or GPU label and temperature
without displaying load percentage. Existing `cpu_temp`, `gpu_temp`,
`cpu_stats`, and `gpu_stats` behavior remains unchanged unless one of the
new options is enabled.

The new options support existing temperature presentation settings,
including Fahrenheit and compact mode. GPU temperature-only display also
supports the existing multi-GPU selection.

### DIFF
--- a/README.md
+++ b/README.md
@@ -365,6 +365,7 @@ Parameters that are enabled by default have to be explicitly disabled. These (cu
 | `cpu_mhz`                          | Show the CPUs current MHz                                                             |
 | `cpu_power`<br>`gpu_power`         | Display CPU/GPU draw in watts                                                         |
 | `cpu_temp`<br>`gpu_temp`<br>`gpu_junction_temp`<br>`gpu_mem_temp`           | Display current CPU/GPU temperature                                                  |
+| `cpu_temp_only`<br>`gpu_temp_only` | Display only the current CPU/GPU temperature                                          |
 | `cpu_custom_temp_sensor`           | Use custom hwmon sensor for cpu temperature. e.g `cpu_custom_temp_sensor=cpuss0_2,temp3_input`.|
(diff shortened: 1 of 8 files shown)


src/hud_elements.cpp
```diff
@@ -281,8 +281,23 @@ void HudElements::version(){
     }
 }
 
+static void display_temp(int temp){
```
drewboardman, line 284: Guide comment: This is the common behavior shared with the existing `gpu_temp` and `cpu_temp`. I've abstracted it to avoid future drift or the requirement to update multiple places.

README.md
```diff
@@ -365,6 +365,7 @@ Parameters that are enabled by default have to be explicitly disabled. These (cu
 | `cpu_mhz`                          | Show the CPUs current MHz                                                             |
 | `cpu_power`<br>`gpu_power`         | Display CPU/GPU draw in watts                                                         |
 | `cpu_temp`<br>`gpu_temp`<br>`gpu_junction_temp`<br>`gpu_mem_temp`           | Display current CPU/GPU temperature                                                  |
```
drewboardman, line 367: IMO this description:

>Display current CPU/GPU temperature

Is inaccurate. I decided not to make "editorial" changes to your README, but this should describe the actual behavior (it includes the % load). If you'd like me to change the wording on the PR I'd be happy to.

